PR #38, opened 2022-09-15
Create Documentation site

e-lo:
This pull request:

- Adds a github action workflow to generate documentation for each branch using mike and mkdocs
- Runs a script in the workflow that auto-generates tables of the frictionless schema (will eventually want to fix some of the formatting here...but it wasn't quick so I held off for now)
- Adds a static (for now) mermaid diagram to the documentation which shows the relationships between tables
- Copies over key markdown files in the base repo to the documentation (i.e. code of conduct, contributing, readme, etc)

NOTE: some troubleshooting may be needed once the GH Actions workflow runs, which wont happen in full until it gets merged to main. 

Closes #22 #23 

image
image
image

PR #39, opened 2022-09-15
Checks for pre-commit and github workflows.

### LINKED ISSUE #25: Add 'checks' workflow for pushes and PRS
As a developer, I'd like to know at a glance if code being pushed:

- [ ] Contains any syntactical errors
- [ ] Conforms to our adopted style guide
- [ ] Passes tests

...and additionally

- [ ] fixes easy-to-fix styling issues (i.e. `black`)

## Implementation Ideas

Could use a `workflow.yml` with pre-commits and a pytest run (if using python)

e-lo:
The purpose of this pull request is to make sure we are committing and merging "good" code and data with consistent formatting. 

This pull request adds both:

1. `pre-commit` checks for developers to check things locally before sending them to github. These are generally pickier than github actions to help with formatting text, etc.
2. Github Workflow to check validation of the frictionless data specs against their json schemas.

NOTE: more linting checks on pushes are defined in #38 

Closes #25


e-lo:
@j-meelah @botanize : I can't get the github action to fail here locally when it should using [`act`](https://github.com/nektos/act) - wondering if either of you could help trouble-shoot in addition to review what I've done so far?

CONTRIBUTING.md
```diff
@@ -1,5 +1,5 @@
-# Contributing to TIDES 
-Thank you for contributing to the TIDES Project.  This document defines the roles and process for 
+# Contributing to TIDES
```
botanize, line 1: pre-commit is telling me that headings need to be surrounded by blank lines. Did you get that warning?

CONTRIBUTING.md
```diff
@@ -1,5 +1,5 @@
...
+Thank you for contributing to the TIDES Project.  This document defines the roles and process for
 contributing to the project and documents the governance roles and approach for decision-making.
 
 ## Contents
```
botanize, line 5: pre-commit is telling me that lists need to be surrounded by blank lines. Did you get that warning?

botanize (changes requested):
- [x] I get a bunch of warnings (blanks-around-lists, blanks-around-headings, heading-increment, no-bare-urls) when I run `pre-commit run --all-files` on your branch. Did you get those warnings as well? If not, is there a config problem?
- [x] I believe the validate_schemas workflow should distinguish between the table schemas and the package schema
- [x] can we drop the language_version requirement from pre-commit black? I don't want to install a two version old python when it works perfectly fine with 3.10
- [ ] I guess there's currently only one dev dependency, pre-commit, which takes care of the rest. If that changes it might make sense to use pyenv and Pipfile instead of requirements.txt? 
- [ ] Add setup instructions for contributors to use these pre-commit hooks

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.8.0
+    hooks:
+      - id: black
+        language_version: python3.8
```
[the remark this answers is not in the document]
e-lo: No good reason :-).  Will remove.

requirements.txt
```diff
@@ -0,0 +1 @@
+pre-commit
```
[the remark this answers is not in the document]
e-lo: I think with one requirement that is a bit overkill? Maybe a todo if things get more involved?

CONTRIBUTING.md
e-lo: Yeah - I fixed it all in PR #38 but will fix here too so it is clean.

CONTRIBUTING.md
e-lo: ditto

.github/workflows/validate_schemas.yml
```diff
@@ -0,0 +1,22 @@
...
+  validate_schemas:
+    name: Validate Frictionless Schemas
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Validate Data Package schema
+        uses: snapcart/json-schema-validator@v1.0.0
+        with:
+          json_schema: "https://specs.frictionlessdata.io/schemas/data-package.json"
+          json_path_pattern: ^spec/.*\.schema.json$
```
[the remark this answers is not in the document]
e-lo: yes!  This is why it wasn't failing when I tested it before. Thanks for catching!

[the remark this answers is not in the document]
e-lo: ok - well I fixed this but it still didn't fix the problem :-/.

setup.cfg
```diff
@@ -0,0 +1,13 @@
+[flake8]
```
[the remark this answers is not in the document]
e-lo: 10-4

e-lo:
@botanize with exception of decision on whether to use a virtual environment at this point - I think I've resolved all your very helpful flags.

BUT - i'm still not getting valid results for the github action workflow using act :-(. It runs cleanly even if I introduce a formatting error :-(

README.md
```diff
@@ -4,19 +4,22 @@ Transit ITS Data Exchange Specification (TIDES) for historical transit operation
...
 
-### TIDES specification
-The TIDES specification is maintained in the `/spec` sub-directory as a series of JSON tables.
+## TIDES specification
 
-Data Table schemas
-Data table schemas are specified in JSON and are compatible with the [frictionless data](https://specs.frictionlessdata.io/table-schema/) table schema standards.
+The TIDES specification is maintained in the `/spec` sub-directory as a series of tables specified as a [frictionless data package](https://specs.frictionlessdata.io/data-package/) defined in `/spec/tides.spec.json`, which is a series of [frictionless data tables](https://specs.frictionlessdata.io/table-schema/) schemas specified in the set of files `/spec/*.schema.json`.
+
+Human readable tables describing the schema can be found at [http://tides-transit/github.io/TIDES/architecture](http://tides-transit/github.io/TIDES/architecture).
```
botanize, line 11: Is this url correct? I think it should be `<http://tides-transit.github.io/TIDES/architecture>`

botanize (changes requested):
I can't run act (no docker license), so I don't have much to say about the validation failing. Check the doc url, but otherwise looks good.

README.md
e-lo: I feel like a doofus, but isn't that what that URL is (in [Markdown format](https://www.markdownguide.org/basic-syntax/#links))

e-lo: or are you saying to just use the shortcut with < >?

botanize: Oh, no, there appears to be a `/` instead of a `.`
```
http://tides-transit/github.io/TIDES/architecture
                    ^
```

e-lo: 🤦‍♀️

e-lo: 👓

botanize (approved):

PR #41, opened 2022-09-19
Fixes for workflows

botanize:
- add token key to jsonschema validator action
 - run validation only on pull-requests, works around <https://github.com/snapcart/json-schema-validator/issues/2>
 - specify python-version for setup-python as '3.x' (latest stable) to use the `cache` feature, see <https://github.com/actions/setup-python/issues/475>
 - use path constraints on validation workflows and split
   - split table spec validation from data package validation to run the correct validation depending on the contents of the PR
   - add `_noop` variants, so that validations succeed if no files match the path constraints (to allow PR merging), see <https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks>

### DIFF
--- a/.github/workflows/validate_package_noop.yml
+++ b/.github/workflows/validate_package_noop.yml
@@ -1,0 +1,14 @@
+name: Validate Frictionless Data Package Schema for Spec
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '!spec/tides.spec.json$'
+
+jobs:
+  validate_data_package:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "no validation required"'
--- a/.github/workflows/validate_package_noop.yml
+++ b/.github/workflows/validate_package_noop.yml
@@ -11,4 +11,4 @@ jobs:
   validate_data_package:
     runs-on: ubuntu-latest
     steps:
-      - run: 'echo "no validation required"'
+      - run: 'echo "no changes to spec/tides.schema.json: no validation required"'
--- a/.github/workflows/validate_package_schema.yml
+++ b/.github/workflows/validate_package_schema.yml
@@ -1,12 +1,14 @@
-name: Validate Frictionless JSON Schemas for Spec
+name: Validate Frictionless Data Package Schema for Spec
 
 on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'spec/tides.spec.json$'
 
 jobs:
-  validate_schemas:
+  validate_data_package:
     name: Validate Frictionless Schemas
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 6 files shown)


.github/workflows/validate_package_schema.yml
```diff
@@ -1,22 +1,23 @@
-name: Validate Frictionless JSON Schemas for Spec
+name: Validate Frictionless Data Package Schema for Spec
 
-on: [push]
+on:
+  pull_request:
+    branches:
+      - main
```
e-lo, line 6: Wouldn't it be helpful to validate for all PRs even if we don't require that it passes?

.github/workflows/validate_package_schema.yml
```diff
@@ -1,22 +1,23 @@
...
-  validate_schemas:
+  validate_data_package:
     name: Validate Frictionless Schemas
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Validate Data Package schema
         uses: snapcart/json-schema-validator@v1.0.0
         with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
e-lo, line 19: oooooHHHHhhh

.github/workflows/validate_package_noop.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches:
+      - main
+    paths:
+      - '!spec/tides.spec.json$'
+
+jobs:
+  validate_data_package:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "no validation required"'
```
e-lo, line 14: add comment on why, i.e.

`no changes to spec/tides.spec.json: no validation required`

.github/workflows/validate_tables_noop.yml
```diff
@@ -0,0 +1,15 @@
...
+      - main
+    paths:
+      - '!spec/*.spec.json$'
+      - '!spec/tides.spec.json$'
+
+jobs:
+  validate_tables:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "no validation required"'
```
e-lo, line 15: add why

e-lo (changes requested):
Tiny comment. TY!!

.github/workflows/validate_package_schema.yml
botanize: This will validate for any PR that targets main, which should cover all cases for changes to the spec.

e-lo: Thinking through "the future" (which maybe we can change this for "the future" when that happens), wouldn't we want it to validate for PRs to any release-candidate type branches as well? I'm just thinking that there might not be a reason to specify a branch at all and if it gets messy/too much we can limit it to certain branches.

botanize: I would be a little more worried about that if we were using the git-flow workflow, but we don't have release branches. I imagine releases will just be tags. The only other branch we would consider now is `ghpages`, but all but one of those files is built from main, and versioning the docs can be done through tags.

I guess maybe there's no big downside to removing the target branch restriction, without release branches it will function the same anyway.

e-lo (approved):

PR #55, opened 2022-09-24
documentation enhancement omnibus PR of 2022

botanize:
This does more than one thing, but they were all connected to improving documentation

- group spec tables by table type (Event, Summary, Supporting)
- sort tables alphabetically
- title case for table titles
- add table description
- add primary key
- add links to GTFS pages referenced
- add links for internal references (foreign IDs)
- add links for rdfTypes

additionally, to make these features function properly, i made the following changes to the spec:

- add a private `_table_type` property (Event, Summary or Supporting)
- fix typo in references to `train_car` to `train_cars`
- change a reference to GTFS trips_performed to an internal reference to the trips_performed table

### DIFF
--- a/docs/document.py
+++ b/docs/document.py
@@ -7,6 +7,8 @@ from typing import Union
 
 import pandas as pd
 
+TABLE_TYPES = ["Event", "Summary", "Supporting"]
+
 
 def _format_cell_by_type(x, header):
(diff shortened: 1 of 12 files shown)


e-lo (approved):
LGTM!
At some point we might want to consider if there are other custom fields we want to add to the schema in addition to `_table_type`

PR #57, opened 2022-09-26
Add .editorconfig

### LINKED ISSUE #54: 🚀💻 – .editorconfig?

As a developer and reviewer I want to have consistent formatting of code across all contributions. [.editorconfig](https://editorconfig.org/) files can configure a wide variety of text editors to function similarly, e.g., trim trailing whitespace, set tabs to 4 spaces for python files, but 2 spaces for yaml, &c.

I'd like a project editor config file, happy to contribute my own.

We could add standard editor settings to the contributing document, but those would not automatically apply.

**Additional context**

botanize:
Uses typical defaults and de facto standards already in place in this repo.

Closes #54

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,6 +15,9 @@ indent_size = 4
 [{.git*,Makefile}]
 indent_style = tab
 
+[{COMMIT_EDITMSG,MERGE_MSG,SQUASH_MSG}]
+max_line_length = 72
+
 [**.md]
 # double whitespace at end of line
 # denotes a line break in Markdown
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,30 @@
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.py]
+indent_size = 4
+
+[{.git*,Makefile}]
+indent_style = tab
+
+[**.md]
+# double whitespace at end of line
+# denotes a line break in Markdown
+trim_trailing_whitespace = false
+x-soft-wrap-text = true
+
+[*.json]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+insert_final_newline = true
+
+[*.py]
+indent_size = 4
+
+[{.git*,Makefile}]
+indent_style = tab
+
+[{COMMIT_EDITMSG,MERGE_MSG,SQUASH_MSG}]
+max_line_length = 72
```
e-lo, line 19: I'd prefer if we use 99. (I know we don't have that yet other places). Thoughts? It seems much more do-able and makes it so you can fit a lot more into in-line comments without having to break them up.

.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+# double whitespace at end of line
+# denotes a line break in Markdown
+trim_trailing_whitespace = false
+x-soft-wrap-text = true
+
+[*.json]
+insert_final_newline = ignore
+
+[*{.ini,.flake8}]
+indent_size = 4
```
e-lo, line 31: Wondering why this isn't 2? (real question...I'm not sure the diff on this vs above)

e-lo (changes requested):
A few minor questions/requests/thoughts.

.editorconfig
botanize: This is just for commit messages. I didn't set a max line length otherwise.

.editorconfig
botanize: Only reason I chose 4 here was that was what was used in the existing files in the repo. I can remove it, and ini/flake8 files will inherit from [*] above (indent_size = 2).

.editorconfig
e-lo: clearly didn't read this close enough

e-lo (approved):
LGTM

PR #62, opened 2022-09-26
WIP: Fix Mike Rebase

### LINKED ISSUE #60: 🐛💻 – mike deploy fails because the remote contains work that you do not have locally
**Describe the problem**

<https://github.com/TIDES-transit/TIDES/actions/runs/3129865893/jobs/5079466105#step:8:25>

```
error: failed to push branch gh-pages to origin: "To https://github.com/TIDES-transit/TIDES
 ! [rejected]        gh-pages -> gh-pages (fetch first)
error: failed to push some refs to 'https://github.com/TIDES-transit/TIDES'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details."
```

I thought the `--rebase` flag was supposed to handle that. Without any versioning, we just keep overwriting `main`, so maybe we want `--ignore` instead of `--rebase`?

The `mike` docs also suggest you might need to manually fetch before running:

```sh
git fetch origin gh-pages --depth=1
```

Maybe we can just change the `fetch-depth` parameter in the `checkout` action to `1`? But I guess that would fetch main, not gh-pages. So probably need to add the `git fetch` command above to the `Build docs` step.

**Steps to reproduce**

e-lo:
This PR resolves the error in #60 where Mike doesn't have the most up to date gh-pages branch on which to rebase.

Fixes #60 

Trying:
 - Fetch GH Pages before mike --rebase

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,9 @@ jobs:
       with:
         fetch-depth: 0
     - name: Document which reference
-      run: echo "Running for ${{ github.ref_name }}"
+      uses: rlespinasse/github-slug-action@v4
+      run: echo "Running for ${{ github.ref_name }}
+           Uses URL Reference ${{ env.GITHUB_REF_SLUG_URL }}"
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
@@ -35,5 +37,5 @@ jobs:
     - name: Build docs
       run: |
         git fetch origin gh-pages --depth=1
-        mike deploy --push --rebase ${{ github.ref_name }}
-        echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ github.ref_name }})"
+        mike deploy --push --rebase ${{ env.GITHUB_REF_SLUG_URL }}
+        echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ env.GITHUB_REF_SLUG_URL }}"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,16 +1,27 @@
-name: Build docs on push
+name: Build and Publish Documentation
 
-on: [push]
+on: 
+  push:
+    paths:
+      - 'docs/*'
+      - '*.md'
+  pull_request:
+    paths:
+      - 'docs/*'
+      - '*.md'
+  workflow_dispatch:
+  create:
 
 jobs:
   build-deploy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - name: Checkout Repo
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
-    - name: Document which branch
-      run: echo ${{ github.ref_name }}
+    - name: Document which reference
+      run: echo "Running for ${{ github.ref_name }}"
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
@@ -31,3 +42,19 @@ jobs:
       run: |
         git fetch origin gh-pages --depth=1
         mike deploy --push --rebase ${{ github.ref_name }}
+    - run: |
+        echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ github.ref_name }})"
+  comment:
+    runs-on: ubuntu-latest
+    if: ${{ github.context.event_name }} == "pull_request"
+    steps:
+      - uses: actions/github-script@v5
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Documentation available at: [`http://tides-transit.github.io/TIDES/${{ github.ref_name }}`](http://tides-transit.github.io/TIDES/${{ github.ref_name }})'
+            })
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,9 +15,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Document which reference
-      uses: rlespinasse/github-slug-action@v4
-      run: echo "Running for ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


e-lo:
@botanize I'm not sure why pushing this didn't trigger a GH action to test it?  🤔

botanize:
Looks like the push target only targets pushes to the TIDES repo in the TIDES-transit org. So pushes to your fork aren't being run. We'll need to add a pull-request target to the workflow to run on pull-requests. I'm ok approving this as is an seeing if it works. But feel free to add a pull-request target to the build docs workflow.

e-lo:
@botanize  - I'm not sure if we can see the effects of this until it is merged to `main`?

.github/workflows/docs.yml
```diff
@@ -1,16 +1,27 @@
...
+  push:
+    paths:
+      - 'docs/*'
+      - '*.md'
+  pull_request:
+    paths:
+      - 'docs/*'
+      - '*.md'
+  workflow_dispatch:
+  create:
```
botanize, line 13: I like the workflow_dispatch trigger, didn't know about that one! I'm not sure about create though. Do we really want to run it for every branch created?
I think the paths restrictions are too strict, they'd need to also include the configuration files, docs/document.py, and the contents of spec/, and now that's pretty much all of the files!

Maybe just 

```yaml
on:
  push:
  pull_request:
  workflow_dispatch:
```

.github/workflows/docs.yml
```diff
@@ -29,4 +40,21 @@ jobs:
         git config --local user.name "github-actions[bot]"
     - name: Build docs
       run: |
+        git fetch origin gh-pages --depth=1
         mike deploy --push --rebase ${{ github.ref_name }}
+    - run: |
```
botanize, line 45: delete this line

.github/workflows/docs.yml
```diff
@@ -29,4 +40,21 @@ jobs:
...
+        echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ github.ref_name }})"
+  comment:
```
botanize, line 47: nice

.github/workflows/docs.yml
```diff
@@ -1,16 +1,27 @@
-name: Build docs on push
+name: Build and Publish Documentation
 
-on: [push]
+on:
```
botanize, line 3: remove trailing whitespace

.github/workflows/docs.yml
```diff
@@ -29,4 +40,21 @@ jobs:
...
+      - uses: actions/github-script@v5
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Documentation available at: [`http://tides-transit.github.io/TIDES/${{ github.ref_name }}`](http://tides-transit.github.io/TIDES/${{ github.ref_name }})'
+            })
```
botanize, line 60: add newline at end of file

botanize (changes requested):
see comments

.github/workflows/docs.yml
e-lo: I dropped the path search but kept 'create' b/c I think we want it to create for branches and tags for now...we can drop if/when it starts being too much.

Note that the delete function will delete the docs for any reference (branch/tag) that we delete from the repo so we shouldn't get too crowded.

botanize (approved):

botanize (changes requested):
There were two workflow failures for this PR, can you take a look at those and see if something needs to change?

e-lo:
Failure Report:

Mike doesn't like versions with non slug-a-ble names (i.e. `merge/62`)

- [ ] Turn GH refs into slugs

GH Action `actions/github-script@v5` got a deprecation warning and other icky stuff that I don't think I can fix quickly (and unsure of ROI). 

- [ ] Delete the triggering job from `workflow.yml`
- [ ] Document as an outstanding issue #69

.github/workflows/docs.yml
```diff
@@ -15,7 +15,9 @@ jobs:
       with:
         fetch-depth: 0
     - name: Document which reference
-      run: echo "Running for ${{ github.ref_name }}"
+      uses: rlespinasse/github-slug-action@v4
+      run: echo "Running for ${{ github.ref_name }}
+           Uses URL Reference ${{ env.GITHUB_REF_SLUG_URL }}"
```
botanize, line 20: I'm seeing an error in the [workflow](https://github.com/TIDES-transit/TIDES/actions/runs/3153704737) that a step can't have both `uses` and `run` keys

.github/workflows/docs.yml
```diff
@@ -35,19 +37,5 @@ jobs:
...
-          github-token: ${{secrets.GITHUB_TOKEN}}
-          script: |
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: 'Documentation available at: [`http://tides-transit.github.io/TIDES/${{ github.ref_name }}`](http://tides-transit.github.io/TIDES/${{ github.ref_name }})'
-            })
+        mike deploy --push --rebase ${{ env.GITHUB_REF_SLUG_URL }}
+        echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ env.GITHUB_REF_SLUG_URL }}"
```
botanize, line 41: missing new line

botanize (changes requested):

botanize:
I was just able to manually update the documentation by first running the "clean", then the "build-deploy" workflows. So maybe this is more of an ordering issue?

e-lo:
@SorenSpicknall If you have time to troubleshoot outstanding issues in this PR it would be great.

PR #64, opened 2022-09-28
disable edit links in docs

### LINKED ISSUE #63: 🐛💻 – edit this page links broken in gh-pages
**Describe the problem**

The edit link for https://tides-transit.github.io/TIDES/main/tables/ is https://github.com/TIDES-transit/TIDES/edit/main/docs/tables.md, but, tables.md is generated by mkdocs/mike and only exists in the gh-pages branch of the repo.

Since these are built from the spec, templates, and some python code, it doesn't really make sense for these pages to have edit links. Editing is done through pull-requests. Can they be disabled for the whole site?

botanize:
closes #63

I had to disable the check-yaml pre-commit hook, due to the following error, which seems totally unrelated:

`could not determine a constructor for the tag 'tag:yaml.org,2002:python/name:mermaid2.fence_mermaid'
  in "mkdocs.yml", line 61, column 19`

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,7 +1,7 @@
 site_name: "Transit ITS Data Exchange (TIDES) Data Specification Suite"
 site_url: https://TIDES-transit.github.io/TIDES
 repo_url: https://github.com/TIDES-transit/TIDES
-edit_uri: edit/main/docs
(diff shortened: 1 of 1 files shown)


e-lo:
See #66 . I think that can supersede this?

PR #72, opened 2022-09-30
Make trip_id_performed optional in vehicle_locations

### LINKED ISSUE #71: 🐛📄 – trip_id_performed should be optional in vehicle_locations
**Describe the problem**
While true AVL systems generally keep track of trips and give then a unique ID, some transit agencies lack these AVL and instead have less expensive GPS fleet tracking units that simply provide a stream of vehicle location events, with vehicle or device ID, a timestamp, and lat/lon coordinates.  `trip_id_performed` is currently marked as required in `vehicle_locations`, so the spec doesn't currently support this use case.  

**Possible Solutions**
Make `trip_id_performed` optional in `vehicle_locations`.  

**Additional context**
Vehicle location heartbeat or breadcrumb data that lacks trips and stops would be processed by a model that would identify trips and stop visits, generating the summary files accordingly, and possibly generating a new version of `vehicle_locations` decorated with the synthesized `trip_id_performed`.

**Process Checklist**

- [x] **Issue Creation:** Fill in required portions of issue template
- [x] **Fact-finding:** Describe potential solutions and recommend one.
- [ ] **Example Updates:** Use samples of real ITS data, solicit samples from other providers and ITS systems.
- [ ] **Consensus:** Build consensus around a single solution.
- [ ] **Implementation:** Implement the change in a pull-request.
- [ ] **Completion:** Follow governance process for pull-request approval.

gabriel-korbato:
Resolves #71

### DIFF
--- a/spec/vehicle_locations.schema.json
+++ b/spec/vehicle_locations.schema.json
@@ -35,10 +35,7 @@
       "name": "trip_id_performed",
       "type": "string",
       "title": "ID referencing trips_performed.trip_id_performed",
-      "description": "Identifies the trip performed.",
-      "constraints": {
-        "required": true
-      }
+      "description": "Identifies the trip performed."
     },
     {
       "name": "stop_sequence",


botanize (commented):
Hi Gabriel,

Thanks for the pull-request. I don't believe we've had sufficient engagement yet to approve of any changes to the spec.

botanize (approved):

PR #75, opened 2022-10-07
Issue 40 validate example data

e-lo:
This pull request:

- adds example data
- updates validation scripts
- adds documentation for adding data and validating it


e-lo:
Issues that are still not working:

1. References in `datapackage.json` are to githubusercontent.com, not local...not sure how to fix?
2. Not sure how to validate a single file using CLI to the schema
3. GH Action?
4. `tides.spec.json` should probably be a [profile](https://specs.frictionlessdata.io/profiles/#introduction)?  

Would love thoughts form @botanize and @j-meelah

botanize:
High-level, I'm not sure about the utility of generating random data to validate. It's a bit circular, and I'm not sure there's any real benefit. For example, could the random data generation script help anyone translate their existing CAD/AVL data into TIDES?

> 1. References in `datapackage.json` are to githubusercontent.com, not local...not sure how to fix?

It doesn't seem like these references are a problem? Though the profile on your datapackage.json should be `tabular-data-package`.

> 2. Not sure how to validate a single file using CLI to the schema

`frictionless validate data/example/data/vehicle_locations --schema spec/vehicle_locations.schema.json`

> 3. GH Action?

For validation? why not just `frictionless validate data/example/data/datapackage.json`? Can even output json or yaml (`--json`, `--yaml`)

> 4. `tides.spec.json` should probably be a [profile](https://specs.frictionlessdata.io/profiles/#introduction)?

I'm not sure anything in the spec matches what we provided in `tides.spec.json`. Everything in the spec that describes a collection requires data resources. Maybe we could make a profile that describes a collection of `tableSchema`? Seems like a lot of work, and the only work tides.spec.json is doing is listing the remaining files in `spec` (`description` is in the table schema, `path` is nonsense, `required` is always false). I recommend removing `tides.spec.json`.

I started producing example data from Metro Transit in the `jpr_sample_data` branch. It's kind of a lot of work to produce! Do you want to look at that and see how we can merge our work?

e-lo:
> It doesn't seem like these references are a problem? 

I was trying to figure out how to make a local reference, which doesn't seem to be as easy in the current structure b/c navigating folders w/out code is...

My goal was to be able to validate the example data against the spec/schemas which were *in that branch* so that we could evaluate any changes to data that any spec changes brought about.

e-lo:
>  I recommend removing tides.spec.json.

I kind of agree, but also think it is useful for doing to auto-documentation generation.  For now, I'll make an issue for us to ponder it.

data/example/scripts/create_example.py
```diff
@@ -0,0 +1,121 @@
...
+        schema_filename (str): Filename with the Frictionless data schema
+        out_dir (str): Where the csv will be written
+    """
+    schema = read_schema(schema_filename)
+    fields = [s["name"] for s in schema["fields"]]
+    schema_name = pathlib.Path(schema_filename).stem.split(".")[0]
+    out_filename = os.path.join(out_dir, schema_name + ".csv")
+    with open(out_filename, "w") as outfile:
+        outfile.write(",".join(fields))
+        print(f"Wrote {out_filename}")
```
botanize, line 103: nit: prefer the unix style, be quiet unless there's a problem

data/example/scripts/requirements.txt
```diff
@@ -0,0 +1,2 @@
+frictionless
+frictionless[pandas]
```
botanize, line 2: still need pandas?

data/README.md
```diff
@@ -0,0 +1,24 @@
+# Data Directory Organization
+
+Each TIDES Data Package example should follow the following directory structure, consistent with the structure of the [Frictionless Data Package specification](https://specs.frictionlessdata.io/data-package/), including:
+
+```
+unique-example-name
+  \data     # data to be validated against the TIDES specification
+  \data\datapackages.json # data package metadata per https://specs.frictionlessdata.io/data-package/
```
botanize, line 8: I prefer this at the top level of the example directory instead of in `data/`, especially if we add custom fields to  `datapackage.json`.

data/README.md
```diff
@@ -0,0 +1,24 @@
+# Data Directory Organization
+
+Each TIDES Data Package example should follow the following directory structure, consistent with the structure of the [Frictionless Data Package specification](https://specs.frictionlessdata.io/data-package/), including:
+
+```
+unique-example-name
+  \data     # data to be validated against the TIDES specification
```
botanize, line 7: I prefer the name `TIDES` for the formatted data directory.

data/README.md
```diff
@@ -0,0 +1,24 @@
+# Data Directory Organization
```
botanize, line 1: I think we need some best practices here. For example, GitHub doesn't love large files and recommends files stay under 50 MB and has a hard limit of 100 MB. GH also recommends repos stay under 1 GB.

So, we should specify that these are **minimal** examples. Maybe aim for 100-1000 lines in any data file, more if absolutely required to reproduce an issue with the spec.

data/example/README.md
```diff
@@ -0,0 +1,16 @@
+# Example TIDES Data Package
+
+Template directory for example scaffolding and helper scripts.
+
+## Scripts for Generating Data
+
+`scripts\create_example.py` has some template code which can help with the following
+
+- `write_schema_examples()`: will generate blank csvs according to the TIDES schema
+- `write_datapackage()` will generate a datapackage.json based on the TIDES schemas and a set of defaults specified in th script.
```
botanize, line 10: I think I'd prefer a `datapackage.json` template instead of a function. Honestly, I'm still trying to figure out what the value of an empty generated example is that wouldn't be better served by a real sample?

main.py
```diff
@@ -0,0 +1,404 @@
...
+        for e in examples:
+            example_title = f"\n## {e.capitalize()}\n\n"
+            example_docs += example_title
+            _readme_filename = os.path.join(data_dir, e, "README.md")
+            if os.path.exists(_readme_filename):
+                example_docs += include_file(_readme_filename, start_line=2)
+            else:
+                example_docs += (
+                    f"No README.md in example folder {os.path.join(data_dir,e)}.\n"
+                )
```
botanize, line 110: This could end up being very large if we have a couple dozen examples with readmes.

How about creating a table and linking to the readme if available? e.g.,

## Examples
| Name | Agency |Vendors| Description | Resources |
|-|-|-|-|-|
| [Example](data/example) | | |[readme](data/example/README.md) | devices, vehicle_locations, fare_transactions, train_cars, operators, stop_visits, vehicle_train_cars, vehicles, trips_performed, station_activities, passenger_events|
| [MetroTransit](data/50027-MetroTransitMN)|50027|TransitMaster, Cubic|| vehicle_locations|

botanize (commented):
This seems like a ton of infrastructure to add a few empty files.

How valuable is it to have a sample that validates no matter what the spec is when the sample is not representative of real data?

I don't love calling the base directory for the samples "data", much prefer "samples".

e-lo:
image
This would be cool if it worked, but it doesn't seem to be (or I can't get it to):

data/example/README.md
e-lo: I don't think the functions serve much purpose other than for developers to be able to auto-generate the example `datapackage.json` and CSV templates.

data/README.md
e-lo: Love this. Will add.

data/example/data/datapackage.json
```diff
@@ -0,0 +1,79 @@
+{
+    "name": "Example TIDES Data Package",
+    "title": "example",
+    "profile": "tabular-data-package",
+    "licenses": [
+        {
+            "name": "Apache-2.0"
+        }
+    ],
```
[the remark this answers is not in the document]
e-lo: 🤔 

Its good practice though for any data package, so to the extent that we have it as something that people are copying, I think it is good to have...

data/example/data/datapackage.json
```diff
@@ -0,0 +1,79 @@
+{
+    "name": "Example TIDES Data Package",
```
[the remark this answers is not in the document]
e-lo: I reversed name and title...fixing

data/example/data/datapackage.json
```diff
@@ -0,0 +1,79 @@
...
+    "profile": "tabular-data-package",
+    "licenses": [
+        {
+            "name": "Apache-2.0"
+        }
+    ],
+    "contributors": [
+        {
+            "title": "My Name",
+            "email": "me@myself.com"
```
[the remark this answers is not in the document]
e-lo: yessss

data/example/data/datapackage.json
```diff
@@ -0,0 +1,79 @@
...
+        }
+    ],
+    "maintainers": [
+        {
+            "title": "Another Name",
+            "email": "another@myself.com"
+        }
+    ],
```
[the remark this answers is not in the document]
e-lo: ✅  NTDID

Wondering if we should add the software to the individual [resources as `sources`](https://specs.frictionlessdata.io/data-package/#metadata)since each resource might have a different one and it might be nice to know which is which (especially when trying to grok somebody elses workflow).

Thoguhts?

data/example/scripts/requirements.txt
e-lo: nope :-)

data/example/README.md
```diff
@@ -0,0 +1,16 @@
+# Example TIDES Data Package
+
+Template directory for example scaffolding and helper scripts.
+
+## Scripts for Generating Data
+
+`scripts\create_example.py` has some template code which can help with the following
+
+- `write_schema_examples()`: will generate blank csvs according to the TIDES schema
```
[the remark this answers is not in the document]
e-lo: I still think there is value, personally. I know I am always doing this for GTFS and wishing that it was documented in some sort of actual schema so I could automagically do it.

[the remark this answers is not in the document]
e-lo: > I don't love calling the base directory for the samples "data", much prefer "samples".
Fixed.  I did this based on recos from repository structures from frictionless but I'm not wedded to it.

data/example/data/datapackage.json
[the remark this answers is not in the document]
e-lo: Per https://specs.frictionlessdata.io/patterns/#private-properties, this shouldn't be a private property b/c it is user specified metadata.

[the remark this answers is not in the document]
e-lo: For now, added:
- `ntd_id` in `datapackage.json`
- `vendor`, `product`, and `component` in `sources` in `resources`

main.py
e-lo: fixed

botanize:
spaces in urls! here's a link that works <https://repository.frictionlessdata.io/pages/dashboard.html?user=TIDES-transit&repo=TIDES&flow=validate+example+tides+data&run=3206719404>

samples/template/TIDES/datapackage.json
```diff
@@ -0,0 +1,171 @@
...
+            "name": "devices",
+            "path": "devices.csv",
+            "schema": "https://raw.githubusercontent.com/TIDES-transit/TIDES/main/spec/devices.schema.json",
+            "sources": [
+                {
+                    "title": "Where did data come from?",
+                    "component": "Type of technology component, i.e. `AVL`",
+                    "product": "Product used.",
+                    "vendor": "Vendor selling product."
+                }
```
botanize, line 37: see previous comment for ideas about content and location of sources

samples/template/scripts/create_template.py
```diff
@@ -0,0 +1,137 @@
...
+import pathlib
+
+
+EXAMPLE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_REPO_DIR = os.path.dirname(os.path.dirname(EXAMPLE_DIR))
+TIDES_SPEC = os.path.join(BASE_REPO_DIR, "spec")
+SCHEMAS = glob.glob(os.path.join(TIDES_SPEC, "**/*.schema.json"), recursive=True)
+SCHEMAS_LOC = "https://raw.githubusercontent.com/TIDES-transit/TIDES/main/spec/"
+
+# DATAPACKAGE.JSON INFORMATION
```
botanize, line 15: Why define a bunch of string and code to write strings to a file? Why to just add a file? Actually, let's take this a step further, and just move the "datapackage.json" text into the readme in "/samples"? Then drop the entire template directory, with the CSVs that I don't think are useful for any purpose.

samples/README.md
```diff
@@ -0,0 +1,83 @@
+# Data Directory Organization
+
+Each TIDES Data Package example should follow the following directory structure, consistent with the structure of the [Frictionless Data Package specification](https://specs.frictionlessdata.io/data-package/), including:
+
+```
+unique-example-name
+  \TIDES     # data to be validated against the TIDES specification
+  \TIDES\datapackages.json # data package metadata per https://specs.frictionlessdata.io/data-package/
```
botanize, line 8: required

README.md
```diff
@@ -6,12 +6,43 @@ This repository provides data schemas and tools to support the access, managemen
...
-The TIDES specification is maintained in the `/spec` sub-directory as a series of JSON tables compatible with the [frictionless data](https://specs.frictionlessdata.io/table-schema/) table schema standards.
+The TIDES specification is maintained in the `/spec` sub-directory as a series of JSON tables compatible with the [frictionless data](https://specs.frictionlessdata.io/table-schema/) table schema and [data package](https://specs.frictionlessdata.io/data-package) standards.
 Human-friendlier documentation is auto-generated and available at:
 
 - [Architecture](architecture.md)
 - [Table Schemas](tables.md)
 
+## Example Data
+
+Example data can be found in the `/data` directory, with one directory for each example.
```
botanize, line 17: change `/data` to `/samples`

README.md
```diff
@@ -6,12 +6,43 @@ This repository provides data schemas and tools to support the access, managemen
...
+TIDES data with a valid [`datapackage.json`](#data-package) can be easily validated using the [frictionless framework](https://framework.frictionlessdata.io/), which can be installed and invoke as follows:
+
+```bash
+pip install frictionless
+frictionless validate path/to/your/datapackage.json
+```
+
+### Data Package
+
+To validate a package of TIDES data, you must add a frictionless-compliant [`datapackage.json`](https://specs.frictionlessdata.io/data-package/) alongside your data which describes which files should be validated to which schemas.  Most of this can be copied from [`/data/template/TIDES/datapackage.json`](https://raw.githubusercontent.com/TIDES-transit/TIDES/main/data/template/TIDES/datapackage.json).
```
botanize, line 30: change link after moving example `datapackage.json` into `/samples/README.md`

botanize (changes requested):
I'm still struggling to see what purpose CSVs with only header rows can serve. With that in mind, I don't really think a "template" directory is very useful at all. I think the simplest and most effective thing to do is to embed an example `dataresource.json` file into `/samples/README.md` and remove all of `/samples/template`.

samples/README.md
```diff
@@ -0,0 +1,83 @@
...
+| `sources` | Array of data sources formatted as a [`source`](#data-source) | Recommended |
+
+### Data Source
+
+| **Field** | **Description** | **Required** |
+| --------- | --------------- | ------------ |
+| `title` | Description of the data source. | Required |
+| `component` | What technology component was used to generate this data (directly or indirectly)? Examples include `AVL`, `APC`, `AFC`, etc.  | Recommended |
+| `product` | What product was used to generate this data (directly or indirectly)? | Recommended |
+| `vendor` | What company makes this product? | Recommended |
```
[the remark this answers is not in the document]
e-lo: The main value I see in putting it under each resource is to explicitly say where each piece of the "TIDES pie" is coming from so that agencies that are trying to mimic another agency's a data flow diagram can do so.  It's not necessarily implicit which resource comes from which product.  For example, there are *many* agencies that receive APC data from their "APC product" that is then manipulated by another product (including internal scripts) and then spat out to other systems.  Exposing this seems like it would add a great deal of value.  That said, there is no reason why you *can't* just put all the sources up top....especially if we aren't articulating a profile which would require it and are merely suggesting.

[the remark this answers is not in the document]
botanize: Sure. It could be nested, big-picture at the top-level, details for each resource, both optional.

samples/template/scripts/create_template.py
```diff
@@ -0,0 +1,137 @@
...
+# Per https://specs.frictionlessdata.io/data-package/
+TITLE = "Template TIDES Data Package Example"
+NAME = "template"
+AGENCY = "Transit Agency Name"
+NTD_ID = "1234-56"
```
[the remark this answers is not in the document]
e-lo: Reduced reporters actually have NTD IDs that look ~  like this , e.g. Humboldt Transit is `9R02-91036`

main.py
```diff
@@ -26,15 +27,46 @@ def define_env(env):
...
+    def list_samples(data_dir: str) -> str:
+        """Outputs a simple list of the directories in a folder in markdown.
+        Args:
+            data_dir (str):directory to search in
+        Returns:
+            str: markdown-formatted list
+        """
+
+        fields = ["Sample", "Agency", "Resources", "Vendors"]
+        df = pd.DataFrame([], columns=fields)
```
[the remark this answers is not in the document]
e-lo: True its a shortcut, but it is prettier.  Happy to update if you feel strongly about it but also feel like pandas is a fine minimum requirement.

samples/template/scripts/create_template.py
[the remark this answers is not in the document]
botanize: No kidding? I've only seen the "5 digit NTD ID" field in the NTD files I've looked at.

main.py
```diff
@@ -133,6 +164,40 @@ def frictionless_schemas(
...
+def sample_metadata_as_dict(dir: str) -> dict:
+    """Reads datapackage and translates key metadata to dictionary.
+
+    Args:
+        dir (str): fully qualified directory for sample data
+
+    Returns:
+        dict: dictionary with data pacakge name, agency, resources and vendors.
+    """
+    dp_filename = os.path.join(dir, "TIDES", "datapackage.json")
```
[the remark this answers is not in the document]
e-lo: (other reason is b/c I'm not sure how well POSIX vs WIN paths are handled...but we'll find out!)

samples/README.md
```diff
@@ -0,0 +1,83 @@
...
+| `profile` | Should be `tabular-data-package` | Required |
+| `licenses` | Should be `[{"name": "Apache-2.0"}]` to be consistent with this repository | Required |
+| `contributors` | Array of data contributors `[{"title": "My Name", "github": "my_handle", "email": "me@myself.com"}]` | Recommended |
+| `maintainers` |  Array of data maintainers `[{"title": "My Name", "github": "my_handle", "email": "me@myself.com"}]` | Recommended |
+| `resources` | Array of data files included in your package, formated as a [`tabular-data-resource`](#data-resource)| Required |
+
+### Template
+
+A `datapackage.json` template is available at [`/data/template/TIDES/datapackage.json`](https://raw.githubusercontent.com/TIDES-transit/TIDES/main/data/template/TIDES/datapackage.json).
+
```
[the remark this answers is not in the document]
e-lo: I don't see why this is better when we can include the file in the markdown? I'd prefer to manage data as data, not as data within markdown.

main.py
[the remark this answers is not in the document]
e-lo: updated.

e-lo:
All, There are two things @botanize are in (what I would hope we both categorize as) healthy disagreement about that it might be useful for the broader community to weigh in on:

1. **Sources**: should we document them at the resource- or datapackage-level? (see discussions above)
2. **Templates**: should we include `datapackage.json` and csv templates that are auto-generated from the spec? Or just include documentation about `datapackage.json` in `README.md`?  Or some middle ground? (see discussions above)

samples/README.md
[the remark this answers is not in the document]
botanize: But it's not data!—it's documentation, an example that doesn't even have meaningful values. Documentation for just about every software tool is [littered](https://docs.opentripplanner.org/en/latest/BuildConfiguration/) with [example](https://en.wikibooks.org/wiki/OpenSSH/Client_Configuration_Files#System-wide_Client_Configuration_Files) [code](https://docs.python.org/3/library/os.html#os.scandir) and [configuration](https://git-scm.com/docs/git-config#EXAMPLES), this is exactly the same.

If you want to include a file, go ahead, but I do think that at the very least the file should have an obviously different name, e.g., `datapackage.json.template` or `datapackage.sample` or something like that so that it's not accidentally interpreted as a real data package description.

e-lo:
Updated this PR to latest `main` and it is passing and I think is ready to merge @botanize ?

README.md
```diff
@@ -41,6 +34,22 @@ Several other validation scripts and tools with more flexibility such as validat
 bin/validate-datapackage [-v remote_spec_ref | -l local_spec_path] [-d dataset_path]
 ```
 
+### Specific files
+
+Specific files can be validated by running the frictionless framework against them and their corresponding schemas as follows:
+
+```sh
+frictionless validate vehicles.csv --schema https://raw.githubusercontent.com/TIDES-transit/TIDES/main/spec/schema.vehicles.json
```
botanize, line 42: This requires `--schema-sync`, or optional fields that are left out will cause validation errors

README.md
```diff
@@ -41,6 +34,22 @@ Several other validation scripts and tools with more flexibility such as validat
...
+```
+
+## Sample Data
+
+[Sample data](https://tides-transit.github.io/TIDES/main/samples) can be found in the `/samples` directory, with one directory for each sample.
+
+### Template
+
+Templates of `datapackage.json` and each TIDES file type are located in the `/samples/template` directory.
```
botanize, line 51: Do you want to mention how these should be used?

bin/validate-data-package
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+# Script: validate_frictionless_package.sh
+# Description: Bash script to validate a Frictionless Data Package using the Frictionless CLI.
+# Usage: validate_frictionless_package.sh [-v tides_version | -l local_schema_location] [-d dataset_location]
```
botanize, line 5: it looks like there's no `.sh` extension on this script, despite what the usage says.

bin/validate-data-package
```diff
@@ -0,0 +1,68 @@
...
+#   -v tides_version: Optional. Specify the version of the TIDES specification or 'local' to
```
botanize, line 6: Is this parameter necessary? Shouldn't this always use the schema specified in the datapackage or treat `-l` as an override? In other words, if `-l` is provided, override the datapackage, otherwise use the datapackage, no need for `-v`.

botanize: In fact, it seems like `-l` should be `-s`, which lets you set any schema location, overriding any datapackage. It wouldn't need to be local, could also be a GitHub location or NFS or samba, &c.

bin/validate-data-package
botanize, line 1: - this file is not marked executable
- how does it differ from bin/validate-datapackage, should it replace it?

bin/validate-data-package
```diff
@@ -0,0 +1,68 @@
...
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      exit 1
+      ;;
+  esac
+done
+
+# Create a temporary data package if using a different schema reference or a local schema
+tmp_datapackage=""
+if [ "$tides_version" != "" ] then
```
botanize, line 39: You're missing a `;` before `then`

bin/validate-data-package
botanize: Also, the samples/template/TIDES doesn't validate, should it?

bin/validate-data-package
```diff
@@ -0,0 +1,68 @@
...
+
+# Update the 'schema' property in the temporary copy of the datapackage.json file, if applicable
+if [ "$tmp_datapackage" != "" ]; then
+  schema_file=$(echo "$tmp_datapackage" | sed 's/\//\\\//g')
+  sed -E -i "s/\"schema\": \"[^\/]+\.schema\.json\"/\"schema\": \"$schema_path_prefix\/\${schema_file##*\/}\"/g" "$tmp_datapackage"
+  dataset_location="$tmp_datapackage"
+fi
+
+# Validate the data package JSON against the TIDES schema
+./validate-data-package-json.sh -v "$tides_version" -d "$dataset_location" -l "$local_schema_location"
```
botanize, line 60: this script does not have a `.sh` extension, nor does it take a `-d` parameter!

bin/validate-data-package
```diff
@@ -0,0 +1,68 @@
...
+
+# Validate the Frictionless Data Package using the Frictionless CLI
+frictionless validate "$dataset_location"
```
botanize, line 63: this might need the `--schema-sync` flag

README.md
```diff
@@ -41,6 +34,22 @@ Several other validation scripts and tools with more flexibility such as validat
 bin/validate-datapackage [-v remote_spec_ref | -l local_spec_path] [-d dataset_path]
```
botanize, line 34: There are 5 scripts in bin/ and the usage information isn't always clear about why you'd use one script or another, in addition, the names are all very similar. I think it would help a lot to document what the purpose of each script is here.

bin/validate-data-package-json
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+
+# Script to validate a local JSON file against a schema specified in a GitHub repository.
+# Usage: validate-data-package-json.sh [-r ref | -l local_schema_location] [-f datapackage_file]
```
botanize, line 4: this script does not have a `.sh` suffix, despite what the usage says

botanize: Attempting to run this script on the sample datapackage.json results in a validation error:

```sh
./bin/validate-data-package-json -l spec -f samples/template/TIDES/datapackage.json           (issue-40-validate-example-data)TIDES
Validating data package file in 
JSON/YAML string must be a a map, cannot load "spec/tides-data-package.json"
```

Am I doing something wrong, or is there a problem somewhere else?

botanize: It also fails without specifying `-l`

```sh
./bin/validate-data-package-json  -f samples/template/TIDES/datapackage.json                  (issue-40-validate-example-data)TIDES
Validating data package file in 
Schema file not found on GitHub for the specified TIDES version: 
```

bin/validate-datapackage-to-profile
botanize, line 1: This script and validate-data-package-json use `-f` but validate-data-package uses `-d`, they should probably use the same parameters for the same inputs

bin/validate-datapackage-to-profile
```diff
@@ -1,16 +1,16 @@
...
 usage="
 Usage: validate-datapackage-to-profile [-r remote_spec_ref | -l local_spec_path] [-f datapackage_file]
    -r remote_spec_ref: Optional. Specify the ref name of the GitHub repository for validating agianst
-             a remote profile where the profile is in the sub-path /spec/tides-data-package.json. 
+             a remote profile where the profile is in the sub-path /spec/tides-data-package.json.
              Should not be used with -l option. Example: -r main
    -l local_spec_path: Optional. Specify the location of the local tides-data-package-json to use.
              Should not be used with -r option. Example: -l spec
-   -d dataset_path: Optional. Specify the path of the datapackage.json file. 
+   -d dataset_path: Optional. Specify the path of the datapackage.json file.
```
botanize, line 13: Usage line says -f, but help and parsing use -d

botanize (changes requested):
Hi E,

I think I've pointed out all the errors and inconsistencies, but you may want to double-check the syntax and usage of the scripts in bin. I also think we need more documentation of those scripts.

I'm also worried that the contents of samples/template/TIDES doesn't validate.

botanize:
@e-lo, can you resolve or address the comments I left in my last review?

chrisyamas:
@e-lo checking in on this one as part of some pre-v2.0 release housekeeping. Obviously it's been open a while and consequently has picked up conflicts with `main` over time. Since the v2.0 release checklist calls for sample datasets covering the new normative changes anyway, it might be good to have the example data and validation rebuilt against v2.0 once it ships rather than rebasing three years of drift here?

But of course I want to make sure we're following the [change management policy](https://tides-transit.org/main/governance/policies/change-management/#versions-and-release-management) here. The release-level language is "Each release SHOULD have sample datasets (real or exemplary) which cover the normative changes in the release," and the [Implementation stage](https://tides-transit.org/main/governance/policies/change-management/#__tabbed_1_5) adds that the TIDES Manager "MUST ensure sample datasets updated or added which support the new feature," which the v2.0 PRs have been handling through the template samples as they go. So if the per-change requirement applies to each PR, and the fuller "real-or-exemplary datasets" goal is at the release-level, then that could be best met by building against the final v2.0 schemas? (if so, that's already on the release checklist in #271).

So I'd like to get you and @botanize to weigh in on how you think we should proceed on this one. Would you rather we close this and open a fresh issue for v2.0 sample data and validation tooling, or revive this branch?

Issue #77, opened 2022-10-08
🚀💻 – resolve frictions in frictionless spec


botanize:
**Describe the feature you want and how it meets your needs or solves a problem**

As a spec writer I want to communicate features of the spec, without creating hidden dependencies or requirements because of strange interpretations within frictionless.

I'm currently running up against two strange design choices in frictionless while trying to create sample data:

1. optional fields in tables are interpreted as required in the file AND nullable. In GTFS an optional (not-required) field is optional in the file AND nullable. I think most people find the GTFS usage more natural. There's currently no way to say a field is optional in a file. This leads to validations failing if an "optional" field is missing, because the frictionless validator requires all fields defined in a schema to be present in the same order they're defined.
2. A foreign key in a table schema suggests that there's an optional link with another table, but in fact, frictionless enforces that the other table exists. For example, `trip_id_performed` in `vehicle_locations` is defined as a foreign key on `trips_performed.trip_id_perfomed`. But that means that if you provide `vehicle_locations.csv` in your data package, you MUST also provide `trips_performed.csv`, `operators.csv` (via trips_performed), `vehicles.csv`, `devices.csv` and `train_cars.csv` (via devices) even if you only provide the required fields, which contain no information not already in `vehicle_locations.csv`.
I'm not even sure you can provide only the required information, for example, if you don't run train service, but provide `vehicle_locations.csv`, the foreign key on `devices` requires you to provide `devices.csv`, but since `devices` also has a foreign key on `train_cars`, you're required to provide `train_cars.csv`, but `train_cars.train_car_id` is required, even though you don't have any train cars, let alone any train car ids that would necessitate an entire `train_cars.csv`. This is clearly not our intention when specifying foreign keys!—I'm totally happy to run some aggregation of `vehicle_locations.csv` on vehicle_id, without knowing anything else about the vehicle.

Unintended foreign key dependencies:

- fare_transactions:
  - vehicles
  - trips_performed
    - operators
  - devices
- passenger_events:
  - vehicles
  - trips_performed
    - operators
  - devices
  - train_cars
- vehicle_locations:
  - vehicles
  - trips_performed
    - operators
  - devices
    - train_cars
- stop_visits:
  - vehicles
  - trips_performed
    - operators
- trips_performed:
  - vehicles
  - operators
- devices:
  - vehicles
  - train_cars
- vehicle_train_cars:
  - vehicles
  - train_cars

**Describe the solution you'd like**

[...]

PR #81, opened 2022-10-27
fix unintended foreign key dependencies

botanize:
This pull request addresses part 2 of issue #77 by using `foreignKeys` instead of `foreignReferences` 

## Context
Using `foreignKeys` in  the spec creates a dependency to another file, which is not the intention. Instead, foreign keys were intended to document optional relationships between files. The existing behavior prevents sample datasets from validating. For example, when providing a `vehicle_locations` file, but not also including `devices` and `train_cars`.

## Options discussed:

- Change the behavior of the validator to change `foreignKeys` from enforced to informational.
- Add an `enforced` attribute to `foreignKeys` so the validator can determine when to enforce the relationship.
- **This PR:** Use a custom property, e.g., `foreignReferences`.

The first two options would require either changes to the existing validator, or forking the validator and maintaining our own. The last option is the easiest, as it requires no changes to the validator, no fork, and has the additional benefit of avoiding language that has a specific and conflicting meaning in databases.

## Decision
Use `foreignReferences` instead of `foreignKeys` in frictionless schema.

## Implementation
The `foreignKeys` property wasn't used anywhere in our documentation pipeline, as of yet, so this is merely a change to the frictionless schemas.

### DIFF
--- a/spec/devices.schema.json
+++ b/spec/devices.schema.json
@@ -62,7 +62,7 @@
       "description": "Indicates the location of a device on the vehicle or at a station. For example:\nFront door.\nBack door.\nEntrance (not located on a vehicle).\nExit (not located on a vehicle).\nEntrance/exit (not located on a vehicle).\nOther."
     }
   ],
-  "foreignKeys": [
+  "foreignReferences": [
     {
       "fields": "stop_id",
       "reference": {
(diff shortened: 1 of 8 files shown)


e-lo (approved):
LGTM!

PR #84, opened 2022-11-01
Support >24 hour service clocks

### LINKED ISSUE #31: 🐛 📄 – timestamp fields of type `time` don't support >24 hour clocks
I need to be able to represent a service date and time like, date: "2022-08-26", time: "24:30:00", or datetime: "2022-08-26 24:30:00", but the frictionless `time` and `datetime` types don't allow that.

## Context
The R [frictionless](https://github.com/frictionlessdata/frictionless-r) package simply removes the time component from `datetime` types that are beyond 24 hours, e.g., '2022-08-26 24:30:00' -> '2022-08-26 00:00:00', and errors on `time` types that are greater than '24:00:00'.  Similarly, the Python [frictionless](https://github.com/frictionlessdata/frictionless-py) package will throw `dateutil.parser._parser.ParserError: hour must be in 0..23: 2022-08-26 24:30:00`.

Many agencies operate service days longer than 24 hours, for example, Metro Transit uses a 28 hour clock. This results in a service date that may not match the calendar date, e.g., a trip operating from 2022-08-26 23:45:00 to 2022-08-27 00:30:00 is all part of the 2022-08-26 service date, but spans the 2022-08-26 and 2022-08-27 calendar dates. In GTFS these times are represented as >24 hours. In this example, the scheduled time of 00:30:00 on 2022-08-27 would be "24:30:00".

The frictionless `datetime` and `time` types however are [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), which is explicitly a 24 hour clock. So the combination of `date` (type = date) and `timestamp` (type = time) fields can not correctly identify the time on the service date. 

For example, for service date 2022-08-26 (Weekday) a trip that starts at 23:45:00 and ends at 24:30:00 can not be correctly identified if the `actual_arrival_time` field is of type `time` because you'd have to choose between having a date + timestamp combination that resulted in a correct calendar time (date: 2022-08-27, timestamp: 00:30:00) vs having the correct service date, but the wrong calendar time (date: 2022-08-26, timestamp: 00:30:00).

## Options

1. Use `datetime` as the type for all time fields that are associated with a service date (date<date>: 2022-08-26, timestamp<datetime>: 2022-08-27 00:30:00). This is a little verbose, but it would be easy for systems to convert service date + time in service day to datetime. For systems without a concept of service date, the `date` and `timestamp` fields would be redundant.
2. Use `integer` as the type for all time fields that are associated with a service date (date<date>: 2022-08-26, time<integer>: 88200). This would be the number of seconds into the service date, e.g., 24:30:00 = 88200. This is often the native timestamp format for systems that use service days, for systems without service days the timestamp could easily be broken down into date + integer time fields. Integer time, while easy for our machine overlords, is not human friendly.
3. Use `string` as the type and `%H:%M:%S` as the format for all time fields that are associated with a service date, same as GTFS (date<date>: 2022-08-26, time<string>: "24:30:00"). This would allow times like 24:30:00 that are not allowed by the ISO 8601 `time` type, and would prevent existing frictionless tools from throwing errors or returning incorrect times when parsing those times. But to turn these strings into times would require a custom parser that would probably end up converting the string to an integer time representation (which has been implemented numerous times for GTFS).

## Decision
Changes all `time` type fields to `datetime` types, and all instances of `time` in the field description to `timestamp`.

botanize:
This PR changes all `time` fields to `datetime`, and all instances of 'time' in the field description to 'timestamp'.

Use of `datetime` instead of `integer` keeps fields human readable and avoids some ambiguities around daylight savings time.

This PR does not rename fields, though that may also be desirable, e.g., `date` to `service_date`, or `timestamp` to `event_timestamp`.

Closes #31

### DIFF
--- a/spec/fare_transactions.schema.json
+++ b/spec/fare_transactions.schema.json
@@ -25,8 +25,8 @@
     },
     {
       "name": "timestamp",
-      "type": "time",
-      "description": "Recorded event time, including for transactions that may be aggregated values associated with a trip or vehicle.",
+      "type": "datetime",
+      "description": "Recorded event timestamp, including for transactions that may be aggregated values associated with a trip or vehicle.",
       "constraints": {
         "required": true
       }
--- a/spec/passenger_events.schema.json
+++ b/spec/passenger_events.schema.json
@@ -25,8 +25,8 @@
     },
     {
       "name": "timestamp",
-      "type": "time",
-      "description": "Recorded event time.",
+      "type": "datetime",
+      "description": "Recorded event timestamp.",
       "constraints": {
         "required": true
       }
--- a/spec/stop_visits.schema.json
+++ b/spec/stop_visits.schema.json
@@ -68,23 +68,23 @@
     },
     {
       "name": "schedule_arrival_time",
-      "type": "time",
-      "description": "Scheduled time at which the vehicle arrives at a stop. References GTFS"
+      "type": "datetime",
+      "description": "Scheduled timestamp at which the vehicle arrives at a stop. References GTFS"
     },
     {
       "name": "schedule_departure_time",
-      "type": "time",
-      "description": "Scheduled time at which the vehicle departs from a stop. References GTFS"
+      "type": "datetime",
+      "description": "Scheduled timestamp at which the vehicle departs from a stop. References GTFS"
     },
     {
       "name": "actual_arrival_time",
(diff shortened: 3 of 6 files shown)


e-lo (approved):
LGTM! Also updated the issue to reflect the decision.

PR #87, opened 2022-11-02
fix odometer type

### LINKED ISSUE #86: 🐛📄 – odometer is type integer, should be number
**Describe the problem**

`vehicle_locations.odometer` is type `integer`, says it references GTFS-realtime, but that is of type `double`, and it seems like sub-meter resolution should be supported.

**Possible Solutions**

Change from `integer` to `number`.

**Additional context**

botanize:
Use `number` type to support sub-meter resolution

Closes #86

### DIFF
--- a/spec/vehicle_locations.schema.json
+++ b/spec/vehicle_locations.schema.json
@@ -131,7 +131,7 @@
     },
     {
       "name": "odometer",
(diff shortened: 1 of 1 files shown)


e-lo (approved):
LGTM

PR #95, opened 2022-12-14
add optional count field

### LINKED ISSUE #73: 📄🚀 – add optional `count` field to passenger_events table

As a developer of data transformation tools, I want a more flexible and potentially more compact representation of event data.

If I understand the passenger_events table correctly, each passenger boarding would be a separate row. Our system does not report individual passenger boarding events, only a single aggregate event. Here's a sample of the rawest APC messages we have access to (some fields removed): 

|     date| vehicle_id|message_timestamp   | passenger_count_on| passenger_count_off|
|--------:|----------:|:-------------------|------------------:|-------------------:|
| 20220930|       4062|2022-09-30 09:22:22 |                  3|                   0|
| 20220930|       3749|2022-09-30 09:22:48 |                  1|                   0|
| 20220930|       2490|2022-09-30 09:22:48 |                  0|                   1|
| 20220930|       3794|2022-09-30 09:37:59 |                  1|                   0|
| 20220930|       2678|2022-09-30 10:41:29 |                  1|                   0|

In the existing spec the first row of the above table would be represented as:

|passenger_event_id|date|timestamp|trip_id_performed|stop_sequence|event_type|vehicle_id|
|-|-|-|-|-|-|-|
|20220930_t1_s1_1|2022-09-30|2022-09-30 09:22:22|t1|1|Passenger boarded|4062|
|20220930_t1_s1_2|2022-09-30|2022-09-30 09:22:22|t1|1|Passenger boarded|4062|
|20220930_t1_s1_3|2022-09-30|2022-09-30 09:22:22|t1|1|Passenger boarded|4062|

The implied count for each row in passenger_events is 1. Let's make that explicit and modifiable with an optional `count` field (positive integer) with a default value of 1. It would be much easier and more compact to represent our passenger events, and be completely backward compatible, e.g.:

|passenger_event_id|date|timestamp|trip_id_performed|stop_sequence|event_type|vehicle_id|count|
|-|-|-|-|-|-|-|-|
|20220930_t1_s1|2022-09-30|2022-09-30 09:22:22|t1|1|Passenger boarded|4062|3|
|20220930_t1_s1|2022-09-30|2022-09-30 09:22:22|t1|1|Ramp was deployed|4062||

Here there's an explicit count provided for the passenger boardings, but none provided for ramp deployed, meaning it happened once.

Alternatively, `count` could be optional but not nullable, so if you provide any value in the `count` field you must provide a value for all rows. That may be a detail left to implementers, simply specifying that the default is 1 if empty should be enough.

This would be useful for not just passenger boardings and alightings, but door opens and other fields, which might happen more than once at a stop, but have a single record in the CAD/AVL system.

Aggregation would change from counting to summing:

```sql
# existing spec: count
SELECT date, trip_id_performed, COUNT(*) AS trip_boardings FROM passenger_events WHERE event_type = 'Passenger boarded' GROUP BY date, trip_id_performed

# proposed spec: using coalesce if the loader doesn't replaces empty counts with 1
SELECT date, trip_id_performed, SUM(COALESCE(count, 1)) AS trip_boardings FROM passenger_events WHERE event_type = 'Passenger boarded' GROUP BY date, trip_id_performed

# proposed spec: using DEFAULT  1 for the count column
SELECT date, trip_id_performed, SUM(count) AS trip_boardings FROM passenger_events WHERE event_type = 'Passenger boarded' GROUP BY date, trip_id_performed
```

Continue to expand aggregated data so that the only difference among a set of rows is the primary key, `passenger_event_id`.

**Additional context and sample data**

botanize:
Adds `count` to the passenger_events table.

I set the minimum to 0, but since the default is 1 and it's not clear what 0 would mean, it should probably be changed to 1.

Closes #73

### DIFF
--- a/spec/passenger_events.schema.json
+++ b/spec/passenger_events.schema.json
@@ -103,6 +103,15 @@
       "type": "string",
       "title": "ID referencing GTFS stops.stop_id",
       "description": "Identifies the stop the vehicle is serving. References GTFS"
+    },
+    {
+      "name": "count",
+      "type": "integer",
+      "title": "Event count",
+      "description": "Count for this event, e.g., 3 for a Passenger Boarding event with 3 boardings, default is  `1`",
+      "constraints": {
+        "minimum": 0
+      }
     }
   ],
   "foreignReferences": [


e-lo (approved):
As discussed during 1/25/23 mtg. LGTM.

PR #96, opened 2022-12-14
add pattern_id

### LINKED ISSUE #90: 🐛📄 – Add pattern and pattern_stop tables
**Describe the problem**
The specification does not model patterns (defined below), but patterns are a key component of transit operations, they are necessary to refer to a stop by sequence within a trip, and they are very useful for aggregation and filtering when generating reports or performing analyses. If a user were to use TIDES with GTFS but without a pattern table, they would need to derive patterns on the fly by distilling them from gtfs.trips. 

**What is a pattern?**

A pattern, also known as a variant or variation, is a sequence of stops,  typically scheduled to be served by one or more vehicle trips. A route is simply a grouping of one or more patterns defined by the transit agency.

Many routes simply have two patterns, one in each direction, while other routes may have additional patterns such as:
- branches
- planned short turns
- planned skipped or added stops (such as during school hours, for express or limited segments, or during temporary closures)
- the same stops but on a different path (for example, dedicated lanes during peak hours, or an express portion on a freeway during part of the day but on a surface street at other times)

**Current work-around**
The current way to work around this is to join trips performed to the schedule and to look up the pattern from the scheduled trip.  But there are issues with this:
1. Not all operated trips are scheduled.  How do we obtain the pattern of an unscheduled trip?
2. GTFS is being promoted as the preferred schedule representation, but it doesn’t model patterns.  (It was designed to display network and schedule information to passengers on a map, not to analyze and manage operations, so patterns are not required.)  GTFS has the shape_id field in the trips table, but that represents a path on a map rather than a sequence of stops.  Some agencies generate shape_id from their scheduling system’s pattern, but this is not a requirement.  The same shape can be used for different patterns (for example, local and limited-stop patterns along the same path).  In this case patterns could be distilled by grouping trips serving the same stops in the same order.
3. Shapes in GTFS don’t have display names that would be useful for generating aggregated reports.  
4. The current approach makes schedules a hard requirement of TIDES, even for analysis that doesn’t involve schedules.  For example, the schedule will always be required to analyze on-time performance, but it shouldn’t be required to analyze running times.  Agencies that don’t have their schedule in GTFS should still be able to use TIDES.

**Possible Solutions**
1. **Add pattern and pattern_stop tables to the TIDES specification.**  The tables would take the following structure:

```
pattern (
    pattern_id      TEXT     NOT NULL
    pattern_name    TEXT
    route_id        TEXT     --foreign key
    direction_id    TEXT     --foreign key
    shape_id        TEXT     --foreign key
    PRIMARY KEY (pattern_id)
)

pattern_stop (
    pattern_id       TEXT        NOT NULL    --foreign key
    stop_sequence    INTEGER     NOT NULL
    stop_id          TEXT        NOT NULL    --foreign key
    cumul_meters     NUMERIC(10,2) --cumulative scheduled distance from first stop
    PRIMARY KEY (pattern_id, stop_sequence)
)
```

The fare_transactions, passenger_events, vehicle_locations, and stop_visits tables would be updated by removing stop_sequence and adding:
- seq_in_pattern,  referencing the pattern_stop table. If GTFS is being used, this field would be a foreign key to GTFS stop_times.stop_sequence.
- seq_in_trip, for the sequence of a visit within its trip.  This may differ from seq_in_pattern if scheduled stops are skipped or if unscheduled visits are made

2. **Suggest adding pattern and pattern_stop tables to the GTFS specification**, and make a GTFS feed a hard requirement of TIDES.  In a sense this would be better because patterns pertain to the schedule, but it may be difficult to modify GTFS, and until they are added to GTFS, TIDES data would be difficult to work with.

3. **Combine both options** by suggesting adding the tables to GTFS, but also adding them to TIDES until they are accepted into GTFS.  If the new tables are never accepted into GTFS, this option is equivalent to the first, with the difference being that we at least tried to add it to GTFS.  In my opinion this is the best option.

4. **Do nothing.**  This is an issue because it makes it difficult to produce aggregate reports by pattern.  See the work-around above and accept its limitations.  In my opinion this is the worst option.

botanize:
Add pattern_id to any file with trip_id_performed.

Closes #90

### DIFF
--- a/spec/fare_transactions.schema.json
+++ b/spec/fare_transactions.schema.json
@@ -75,6 +75,12 @@
       "description": "Identifies the trip performed. May be null if the fare collection device is NOT located on a vehicle. May be null if on a vehicle but trip-level data is unavailable, in which case the data would be associated with the vehicle."
     },
     {
+      "name": "pattern_id",
+      "type": "string",
+      "title": "unique identifier for the route-pattern",
+      "description": "Identifies the unique stop-path for a trip, may be distinct from GTFS shapes.shape_id"
+    },
+    {
       "name": "stop_sequence",
       "type": "integer",
       "title": "positive integer referencing GTFS stop_times.stop_sequence",
--- a/spec/passenger_events.schema.json
+++ b/spec/passenger_events.schema.json
@@ -103,6 +103,12 @@
       "type": "string",
       "title": "ID referencing GTFS stops.stop_id",
       "description": "Identifies the stop the vehicle is serving. References GTFS"
+    },
+    {
+      "name": "pattern_id",
+      "type": "string",
+      "title": "unique identifier for the route-pattern",
+      "description": "Identifies the unique stop-path for a trip, may be distinct from GTFS shapes.shape_id"
     }
   ],
   "foreignReferences": [
--- a/spec/stop_visits.schema.json
+++ b/spec/stop_visits.schema.json
@@ -39,6 +39,12 @@
       }
     },
     {
+      "name": "pattern_id",
+      "type": "string",
+      "title": "unique identifier for the route-pattern",
+      "description": "Identifies the unique stop-path for a trip, may be distinct from GTFS shapes.shape_id"
+    },
+    {
       "name": "vehicle_id",
       "type": "string",
       "title": "ID referencing vehicles.vehicle_id",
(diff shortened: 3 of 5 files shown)


e-lo (approved):
As discussed in 1/25/23 mtg - LGTM

Issue #40, opened 2022-09-15
🚀💻 – Validate sample data against TIDES frictionless specification


e-lo:
# User Stories
**Describe the feature you want and how it meets your needs or solves a problem**

1. As a transit agency, I'd like to know how to add my sample data.
2. As a transit agency, I want to know if my data and the scripts that support them conforms to the latests TIDES spec so that I can make any necessary fixes.
3. As a TIDES contributor, I'd like to know how proposed spec changes affect actual data so I can evaluate their ROI.
4. As a transit agency or transit technology developer, I'd like to easily review the sample data, scripts, and the context in which they were developed.
5. As a smaller transit agency or one that isn't yet automatically generating data, I'd like a template folder that can be used for quickly prototyping my TIDES data.
6. As a TIDES maintainer, I'd like templates to automatically update based on updates to the spec

# Proposed Solution

List of Solutions for (relevant user story)

(Checking box indicates consensus achieved on approach)

- [x] Documentation which articulates a desired example folder structure (1)
- [x] GH Workflow for [validation of /TIDES folders](https://repository.frictionlessdata.io/) (2,4)
- [x] Automatic documentation generation for example data (4) 
- [ ] Fill template directory `datapackage.json` data with REPLACEME (or similar) 
- [ ] Template folder with required and suggested directories and files (including `datapackage.json`) which can be used to quickly generate new examples (5)
- [ ] Auto-generation of templates based on updates to the spec (6)

BONUS (or potentially another issue/PR):

- Issue generation for failed validations

Proposed example directory structure: 

```
/samples
    /agency-name     # Unique agency name
        datapackage.json  # Basic example information such as agency-name, CAD-AVL vendor, spec version, data maintainer (and their GH handle) 
        /TIDES       # Data formatted in TIDES standard (in the future, we could have subfolders for versions if necessary)
        /raw    # Raw input data
        /scripts     # Scripts that turn data-raw to TIDES
```

# Consensus Building

## General Agreement

- User stories 1-4 are useful and their proposed implementation in PR generally agreed upon

### 1 - Sources

*How should data sources be documented?*

#### Context

- We are interested in understanding/daylighting/documenting the products or components the data came from.
- `datapackage.json` allows users to specify where the data came from in the [`sources`](https://specs.frictionlessdata.io/data-package/#metadata) field.
[...]

PR #100, opened 2023-01-07
Add documentation of sample data structure

e-lo:
# Pull Request

This PR adds documentation of desired samples structure and generation for sample. 

This reflects the splitting up of the original PR reflecting Issue #40 (PR [#75](https://github.com/TIDES-transit/TIDES/pull/75)) per the [suggestion made by](https://github.com/TIDES-transit/TIDES/issues/40#issuecomment-1320525834) @botanize and includes:


samples/README.md
```diff
@@ -0,0 +1,74 @@
...
+3. *Include a README.md* in the base folder of your example with an overview so that it can be included in the documentation.
+
+## Data Package
+
+TIDES data packages must include a [`datapackage.json`](https://specs.frictionlessdata.io/data-package/).  Key information to include in [`datapackage.json`](https://specs.frictionlessdata.io/data-package/) includes:
+
+| **Field** | **Description** | **Required** |
+| --------- | --------------- | ------------ |
+| `title` | A human-readable title. | Required |
+| `name` | Identifier string as a URL-friendly slug. | Required |
```
botanize, line 28: Having both `title` and `name` is the most confusing thing to me about the sample data spec. `name` is basically the title, but has the additional property that it must be a url-friendly slug, which is not really defined here. The data package spec only recommends a `name`, `resources` is the only required property. Maybe we can add an explanation of why we require a url-slug and what url-slug means? If we don't intend to use name, because we're not creating a data portal (in the frictionless sense), maybe we can make it optional or remove it?

- name must contain only: lowercase alpha-numeric ASCII characters, underscores (_), en-dashes (-), and periods (.)

samples/README.md
```diff
@@ -0,0 +1,74 @@
...
+pip install frictionless
+frictionless validate path/to/your/datapackage.json
+```
+
+### Specific files
+
+Specific files can be validated by running the frictionless framework against them and their corresponding schemas as follows:
+
+```sh
+frictionless validate vehicles.csv --schema https://raw.githubusercontent.com/TIDES-transit/TIDES/main/spec/vehicles.schema.json
```
botanize, line 72: need `--schema-sync` flag here or csv header is ignored and fields are assumed to be in the exact position as defined in the schema, essentially making all fields required.

botanize (commented):
I see a few places we could make some nice usability improvements, let me know what you think about my suggestions.

README.md
```diff
@@ -6,12 +6,43 @@ This repository provides data schemas and tools to support the access, managemen
...
+```sh
+frictionless validate datapackage.json
+```
+
+### Specific files
+
+Specific files can be validated by running the frictionless framework against them and their corresponding schemas as follows:
+
+```sh
+frictionless validate vehicles.csv --schema https://raw.githubusercontent.com/TIDES-transit/TIDES/main/spec/schema.vehicles.json
```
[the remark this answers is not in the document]
e-lo: fixed!

samples/README.md
e-lo: I added a link to explain slugs and changed the description slightly. I also made it recommended instead of required.

samples/README.md
```diff
@@ -0,0 +1,74 @@
...
+| `vendor` | What company makes this product? | Recommended |
+
+## Data validation
+
+Data with a valid [`datapackage.json`](#data-package) can be easily validated using the [frictionless framework](https://framework.frictionlessdata.io/), which can be installed and invoke as follows:
+
+```bash
+pip install frictionless
+frictionless validate path/to/your/datapackage.json
+```
```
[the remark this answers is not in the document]
e-lo: ahhh - i thought you were talking before about it SHOULD only validate the datapackage file and not the resourcs.  Will investigate.

main.py
```diff
@@ -133,6 +160,49 @@ def frictionless_schemas(
...
+    Returns:
+        str: row in a markdown table with name, agency, resources, vendors
+    """
+    dp_filename = os.path.join(dir, "datapackage.json")
+    if not os.path.isfile(dp_filename):
+        raise ValueError(f"Can't find datapackage file:\n   {dp_filename}.")
+    with open(dp_filename, "r") as dp_file:
+        dp = json.loads(dp_file.read())
+
+        _vendors = [
```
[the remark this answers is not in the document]
e-lo: addressed/fixed

samples/README.md
```diff
@@ -41,7 +41,7 @@ Key fields for each [`tabular-data-resource`](https://specs.frictionlessdata.io/
 
 | **Field** | **Description** | **Required** |
 | --------- | --------------- | ------------ |
-| `name` | Short sluggable name used to refer to data in this file. | Required |
+| `name` | Short [sluggable](https://en.wikipedia.org/wiki/Clean_URL#Slug) name used to refer to data in this file. `name` must be unique within this datapackage. | Required |
```
botanize, line 44: Is this sluggable, or does it need to be a slug?

samples/README.md
```diff
@@ -25,7 +25,7 @@ TIDES data packages must include a [`datapackage.json`](https://specs.frictionle
 | **Field** | **Description** | **Required** |
 | --------- | --------------- | ------------ |
 | `title` | A human-readable title. | Required |
-| `name` | Identifier string as a URL-friendly slug. | Required |
+| `name` | Short identifier string as a [URL-friendly slug](https://en.wikipedia.org/wiki/Clean_URL#Slug). | Recommended |
```
botanize, line 28: Same question as line 44 below, plus these descriptions should match

main.py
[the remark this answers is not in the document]
botanize: It doesn't look like the fix matches the example data structure I provided, and there's no documentation in this PR for the higher-level source information.

botanize (changes requested):
a couple of consistency issues

botanize:
Big question, should the samples live in their own repo, e.g., TIDES-samples? We could still load them in for testing as a git submodule. It would make regular contributions and maintenance on the TIDES-transit repo slightly easier by avoiding large data transfers with pulls.

e-lo:
> Big question, should the samples live in their own repo, e.g., TIDES-samples? 
I would say - let's start simpler (this PR which we have been discussing for months) and move to more complex if/when we need to.

main.py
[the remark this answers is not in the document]
e-lo: added documentation to samples/readme and updated code to support nesting of vendor after sources

samples/README.md
e-lo: should be sluggable.

samples/README.md
e-lo: Made more consistent.

SorenSpicknall (approved):
Approving in its current state, though we'll still want to follow up on the issue with validation discussed in the PR conversation.

e-lo:
Closing because replaced by #162

PR #104, opened 2023-02-01
Add github workflow to validate samples

e-lo:
This PR adds a GitHub workflow for sample validation and adds a script to locally validate schemas and samples as well as lint.

Also:
- linting
- added requirements for `jsonschema`
- updated documentation


e-lo:
While it doesn't seem like there is the ability to add `--schema-sync` flag to the frictionless repository action at this time, [they have created an issue for it](https://github.com/frictionlessdata/repository/issues/38).

You *can* at this time add an `inquiry` to the repository action, which in turn can skip various errors such as [`extra-label`](https://framework.frictionlessdata.io/docs/errors/label.html).  That said, I don't see a way to make sure that optional fields in the schema aren't flagged...

botanize (commented):
Ok, glad they're looking into it. I don't think we should merge this workflow until there's a mechanism to specify `--schema-sync`, otherwise it will just produce huge reports (1 line per row of the csv!) full of spurious errors.

The alternative is to run `frictionless validate --schema-sync` directly in the workflow instead of using the frictionless `repository` workflow.

test
```diff
@@ -0,0 +1,24 @@
...
+echo "2/5 Validating table schemas"
+pre-commit run --all-files
+echo "Processing file: $file"
+for file in spec/*schema.json; do
+  echo "Validating: $file"
+  frictionless validate $file
+done
+
+echo "3/5 Validating datapackage"
+json-schema-cli validate https://specs.frictionlessdata.io/schemas/data-package.json spec/datapackage.json
```
botanize, line 15: I see jsonschema-cli is now part of requirements.txt, but this script (and the documentation) doesn't do anything to  make sure it's installed. Should we add `pip install -r requirements.txt` to the top of this script? Should this be done in a virtualenv? Should we just update the documentation to say "install requirements with `pip install -r requirements.txt`"?

The requirements installs a surprising number of packages, so I guess I'd prefer that this script were run in a virtual environment, with requirements installed into the venv only. Options include:

- python3 -m venv .env
- [rye](https://github.com/mitsuhiko/rye)
- [pipenv](https://pipenv.pypa.io/en/latest/index.html)

botanize, line 15: This should be `jsonschema-cli`, but even then I get an error. Looks like spec/datapackage.json doesn't exist and should instead be `spec/tides-data-package.json`, but I still get the following error.
```
% jsonschema-cli validate https://specs.frictionlessdata.io/schemas/data-package.json spec/tides-data-package.json
JSON/YAML string must be a a map, cannot load "https://specs.frictionlessdata.io/schemas/data-package.json"
```

botanize (changes requested):
Hi! Overall this is good, but a few specific issues that need to be dealt with.
- [x] move away from deprecated `set-output` command
- [x] deal with dependencies for `test` script
- [x] fix jsonschema-cli validation step

See other comments for suggestions.

Joey

test
e-lo: thx for flagging!  It was too silent a fail for me to notice in my multitasking.  I updated the script to download the profile first (although frustrating that it doesn't have that feature baked in) and made the whole locally testing script/flow more legible so that info-only is suppressed and errors are explicit.

e-lo: We already had instructions int the setup to install the requirements. I added a recommendation to install into a virtual environment with these recommendations but don't think we want to be forceful about installing them (and how and leave that up to the user. 
image

.github/workflows/validate_samples.yml
```diff
@@ -0,0 +1,72 @@
...
+          samples=$(find ./samples -type d -name 'TIDES')
+          echo "::set-output name=samples::$samples"
+      - name: Validate Sample Data
+        id: validation
+        run: |
+          samples=(${{ steps.sample_folders.outputs.samples }})
+          results=""
+          for sample in "${samples[@]}"; do
+            result=$(frictionless validate "$sample/datapackage.json" --json)
+            results+="$result\n"
```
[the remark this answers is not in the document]
e-lo: This *should* upload the result as an artifact...but it is hard to test w/out example data.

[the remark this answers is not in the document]
e-lo: >...validate only the changed sample data for a sample data push.

I think we want it to validate all the files each time since if the failure comes from a spec change that is also important to know.

.github/workflows/validate_samples.yml
```diff
@@ -0,0 +1,72 @@
...
+          pwd
+          samples=$(find ./samples -type d -name 'TIDES')
+          echo "::set-output name=samples::$samples"
+      - name: Validate Sample Data
+        id: validation
+        run: |
+          samples=(${{ steps.sample_folders.outputs.samples }})
+          results=""
+          for sample in "${samples[@]}"; do
+            result=$(frictionless validate "$sample/datapackage.json" --json)
```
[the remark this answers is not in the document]
e-lo: yes! the whole reason why we did this. Added.

.github/workflows/validate_samples.yml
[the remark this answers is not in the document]
botanize: My suggestion was to validate all sample data when the spec is changes, but validate only the changed sample data when a change to sample data is what triggers the workflow.

.github/workflows/validate_samples.yml
```diff
@@ -0,0 +1,72 @@
...
+            results+="$result\n"
+          done
+          echo "::set-output name=results::$results"
```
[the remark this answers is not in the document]
e-lo: 10-4, fixed

.github/workflows/validate_samples.yml
```diff
@@ -0,0 +1,72 @@
...
+            for (let i = 0; i < samples.length; i++) {
+              const sample = samples[i];
+              const result = JSON.parse(results[i]);
+              let status = '';
+              if (result.valid) {
+                status = '![passing](https://via.placeholder.com/20/00ff00/000000?text=+)';
+              } else if (result.stats.errorCount > 0) {
+                status = '![failing](https://via.placeholder.com/20/ff0000/000000?text=+)';
+              } else {
+                status = '![warning](https://via.placeholder.com/20/ffff00/000000?text=+)';
```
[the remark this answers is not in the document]
e-lo: We can do that - benefit of using the service is that we can reference it in readme files, etc.

tests/validate_samples
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+echo "Validating sample data"
+for file in samples/**/TIDES/datapackage.json; do
+    echo "Validating: $file"
+    frictionless validate --schema-sync $file
```
botanize, line 5: I think `**` should be `*`, as is the validation fails (with no samples folder) and the output of echo is that `samples/**/TIDES/datapackage.json` is considered a file.

tests/test_all
```diff
@@ -0,0 +1,47 @@
...
+pre-commit run --all-files
+
+# Iterate over script files in the directory
+for script_file in "${script_files[@]}"; do
+
+  # Exclude the current script and non-executable files from execution
+  if [ "$(basename "$script_file")" != "test_all" ]; then
+    if [ -x "$script_file" ]; then
+      # Update progress
+      echo "----- Running test $((i + 1)) of $num_files: $script_file"
```
botanize, line 22: `i` doesn't seem to be updated, so all of my "------ Running" lines start with `----- Running test 1 of 5` and I never see `2 of 5`.

tests/test_all
```diff
@@ -0,0 +1,47 @@
...
+        echo "$output" >&2
+      fi
+    else
+      # Notify non-executable files
+      echo "!!! File is not executable so was not run: $script_file."
+      echo
+    fi
+  fi
+done
+echo "----- Running test $num_files/$num_file: Building documentation locally"
```
botanize, line 39: There is no `$num_file`, should be `$num_files`.

tests/validate_samples
botanize: Also, the glob by default returns something if there's no match, need to set nullglob to not run if there's no sample data, e.g.,
```
#!/usr/bin/env bash
echo "Validating sample data"
shopt -s nullglob
for file in samples/*/TIDES/datapackage.json; do
    echo "Validating: $file"
    frictionless validate --schema-sync "$file"
done
shopt -u nullglob
```

botanize (changes requested):
Some lingering errors in tests

.github/workflows/validate_samples.yml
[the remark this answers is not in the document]
e-lo: right now there is zero sample data so i think we can add this when it is relevant

.github/workflows/validate_samples.yml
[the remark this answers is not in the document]
e-lo: updated to use acticons

tests/test_all
e-lo: good catch - i changed how the looping worked and forgot to update this part...fixing now

tests/test_all
e-lo: good catch - didn't come up for me b/c it was in my env from before :-(

e-lo: fixed

e-lo:
@botanize I think i got all your comments addressed in subsequent commits minus the one suggesting that only do validation on the files that actually changed (we can address that when we have files to do).

botanize (approved):
I'd still like to test this on actual sample data, but everything else seems to work, so let's merge.

PR #116, opened 2023-03-28
Fixed fields clashing with SQL keywords

### LINKED ISSUE #82: 🐛📄 – Some fields in the TIDES spec clash with Protected Keywords in SQL Databases
**Describe the problem**
While starting to sketch out the development of new data tables to store data that complies with the TIDES spec, we noticed several of the column names clash with protected SQL keywords. One example is in `vehicle_locations.schema.json`, where the service date column is named `date`, which is protected in most relational databases.

**Possible Solutions**

* Define a standard for changing the name of reserved words (e.g. `date` becomes `dt` instead when it is a field name) or require more detail in proposed field names (e.g. `date` becomes `service_date` in `vehicle_locations.schema.json`)
* Adjust the CONTRIBUTING.md file to specify part of the code review process will check against reserved words in common SQL databases
* Instead, could accept that the spec includes protected keywords (since it would be difficult to create a list of even the more common programming languages beyond just SQL), and put the onus on data owners and tool creators to define required workarounds.

**Additional context**
Here is a list of reserved keywords from most major database systems that could potentially serve as a filter for the new spec: [Drupal - List of SQL Reserved Words](https://www.drupal.org/docs/develop/coding-standards/list-of-sql-reserved-words)

Danishswag:
Updates several field names to avoid SQL keywords in common databases:

Schema | Current Field Name | Proposed Field Name
-- | -- | --
fare_transactions | date | service_date
fare_transactions | timestamp | event_timestamp
passenger_events | date | service_date
passenger_events | timestamp | event_timestamp
passenger_events | count | event_count
vehicle_locations | date | service_date
vehicle_locations | timestamp | event_timestamp
station_activities | date | service_date
stop_visits | date | service_date
stop_visits | checkpoint | timepoint
stop_visits | load | vehicle_load
trips_performed | date | service_date
devices | location | device_location
vehicle_train_cars | order | train_car_order

Closes #82

### DIFF
--- a/spec/stop_visits.schema.json
+++ b/spec/stop_visits.schema.json
@@ -12,7 +12,7 @@
   ],
   "fields": [
     {
-      "name": "date",
+      "name": "service_date",
       "type": "date",
       "description": "Service date. References GTFS indirectly via calendars.txt and calendar_dates.txt",
       "constraints": {
@@ -68,7 +68,7 @@
       "description": "Identifies the stop. References GTFS"
     },
     {
-      "name": "checkpoint",
+      "name": "timepoint",
       "type": "boolean",
       "description": "Indicates if the stop should be used for evaluating schedule adherence, on-time performance, and other KPIs. This could be populated to match the GTFS “timepoint” field."
     },
@@ -133,7 +133,7 @@
       }
     },
     {
-      "name": "load",
+      "name": "vehicle_load",
       "type": "integer",
       "description": "Number of riders on the vehicle when departing the stop.",
       "constraints": {
--- a/spec/stop_visits.schema.json
+++ b/spec/stop_visits.schema.json
@@ -133,7 +133,7 @@
       }
     },
     {
-      "name": "vehicle_load",
+      "name": "departure_load",
       "type": "integer",
       "description": "Number of riders on the vehicle when departing the stop.",
       "constraints": {
--- a/spec/devices.schema.json
+++ b/spec/devices.schema.json
@@ -57,7 +57,7 @@
       "description": "Model of the device as specified by the vendor."
     },
     {
-      "name": "location",
+      "name": "device_location",
       "type": "string",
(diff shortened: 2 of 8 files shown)


spec/stop_visits.schema.json
```diff
@@ -133,7 +133,7 @@
       }
     },
     {
-      "name": "load",
+      "name": "vehicle_load",
```
botanize, line 136: I'm ok with vehicle load, but wonder if "passenger_load" might be more appropriate

botanize (approved):
Hi, I think we can use this as is, but wonder if anyone else has any feelings about vehicle_load vs passenger_load.

spec/stop_visits.schema.json
gabriel-korbato: I would be OK with both.  `vehicle_load` is useful to distinguish from platform or escalator loads, all of which are passenger loads, but in the context of the stop_visits table it's clear that passenger_load would refer to the load in the vehicle.  What isn't clear is if the load is measured upon arriving at the stop or after departing it.  If we want the column name to be self-documented we could consider `departure_load` or `arrival_load`.

botanize: +1 for `departure_load`

jaygordon: I prefer `departure_load` too, or if we want shorter, easier to type names, `load_out`, which I admit is less clear.

Danishswag: Hi all, I pulled `vehicle_load` from the field description, but definitely support other options. For `departure_load`, is it potentially confusing given that some systems will transmit an arrival load at the terminal stop? Happy if we define it specifically as departure load too (so you know arrival load at the terminal would be departure load at the second to last stop).

botanize: Good point. The description should probably be updated to define this as the load at arrival for the final stop of a trip and departure for all others. Are there systems that provide only arrival loads? Do we need `arrival_load` and `departure_load` fields?

botanize: Sounds like the simplest answer is to use `departure_load` and document that it means the load at departure from the stop in the description. I agree that converting between arrival and departure load is pretty trivial, so lets just leave that out, since I don't think the description is a good place for it.

botanize (approved):

PR #129, opened 2023-04-20
Fix mermaid

botanize:
Pins mermaid to version 9.4.3, to avoid incompatibility between version 10 and mkdocs. 

There's an open issue to make version 10 work with mkdocs.
https://github.com/fralau/mkdocs-mermaid2-plugin/issues/70

There's a work-around in the above issue where you create a new javascript file that loads the ESM module of mermaid v10. I'm open to that as well.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -43,7 +43,7 @@ watch:
   - README.md
 
 extra_javascript:
-  - https://unpkg.com/mermaid/dist/mermaid.min.js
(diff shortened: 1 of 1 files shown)


botanize:
preview the fix here: https://tides-transit.github.io/TIDES/fix_mermaid/architecture/

e-lo:
LGTM

e-lo (approved):

Issue #50, opened 2022-09-23
🐛📄 – Valid IDs may not exist in GTFS


botanize:
**Describe the problem**

If MetroTransit creates a new trip with a new trip_id to fill a service gap, that trip_id won't appear in GTFS trips.txt. None of the additional information provided by GTFS (service_id, route_id, direction_id, scheduled times and distances) would be available for trips with IDs that don't match the static GTFS. This would prevent aggregation of event data, and inclusion in KPIs.

For example, if I insert a new trip into a 40 minute headway gap, the new trip_id won't be in GTFS, the headway KPI won't be able to identify that trip as belonging to the same route-direction as the surrounding trips, so the calculated headways will be 40 instead of 20 minutes.

**Possible Solutions**

- Add supplementary tables for service adjustments, following GTFS-realtime Service Changes: new_routes, new_trips, new_shapes, new_stops. This would make the content of service adjustment available to TIDES for matching and aggregation, but we'd be replicating parts of GTFS-realtime ServiceChanges into TIDES.

**Additional context**

gabriel-korbato:
TIDES allows operated trips to have a null scheduled trip ID.  It is the scheduled trip that points to GTFS, not the operated trip.  Aggregation should happen across operated trips, so it shouldn't matter if the trip appears in GTFS.  The trips_performed table has route_id, shape_id, and direction_id, which could be populated even if the trip wasn't scheduled.  This is why those fields were included in that table.

botanize:
Ah, sure. However, a new `shape_id` (that doesn't exist in GTFS static) means I don't have access to `shape_dist_traveled` or even the shape geometry with which I could calculate scheduled distances. A new `route_id` means I may not have a `route_short_name`/`route_long_name` pair available, nice when your `route_id` is 901 but your route name is "METRO Blue Line". New `trip_id`s mean no `trip_headsign`, `block_id`, &c.

Maybe the answer is just, if you need extended properties, you need to supply them yourself. This could be the basis for a detailed best practice.

e-lo:
Also to consider is GTFS RT TripDesriptor and their uniqueness requirements should a `trip_id` not be present in published GTFS Schedule...

For a frequency-based schedule where there is one `trip_id` but many trips, they should be uniquely defined w/
- `trip_id`
- `start_time` (which should be stable...even if it isn't true!?)
- `start_date`

Alternatively:

- `route_id`
- `direction_id`
[...]

PR #135, opened 2023-05-03
Adds description about consistency of trip_ids

e-lo:
This PR adds explicit description of the implied relationship between `trip_id` in GTFS and `trip_id_performed` if and when trips exist in both datasets.

Based on discussion for issue #50

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -82,3 +82,9 @@ graph LR;
     click passenger_events "../tables/#passenger-events"
     click fare_transactions "../tables/#fare-transactions"
 ```
+
+!!! tip "Relationship of `trip_id_performed` to GTFS Schedule `trip_id`"
+
+    `trip_id_performed` in TIDES and `trip_id` in GTFS Schedule should be consistent when a performed trip exists in the published GTFS Schedule. 
+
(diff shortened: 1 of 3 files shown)


docs/architecture.md
```diff
@@ -82,3 +82,9 @@ graph LR;
     click passenger_events "../tables/#passenger-events"
     click fare_transactions "../tables/#fare-transactions"
 ```
+
+!!! tip "Relationship of `trip_id_scheduled` to GTFS Schedule `trip_id`"
+
+    `trip_id_scheduled` in TIDES and `trip_id` in GTFS Schedule should be consistent when a performed trip exists in the published GTFS Schedule. 
+
+    If a trip does not exist in the published GTFS Schedule, `trip_id` values in TIDES do not have any assumed relationship.
```
gabriel-korbato, line 90: Do you mean `trip_id_scheduled` values in TIDES do not have any assumed relationship?  And does that mean that they are blank/null?

Separately, if an agency is using scheudule data not in GTFS format, `trip_id_scheduled` might reference that.  So should we say "If a trip does not exist in the published GTFS (or other) schedule, ..."

e-lo: Thanks for catching! I think that since this refers to GTFS specifically that is a bit moot. I added a clarification that " any relationship...with GTFS"

PR #144, opened 2023-06-15
Update trips_performed.schema.json

### LINKED ISSUE #51: 🐛📄 – trips_performed primary key {date, trip_id} is insufficient to identify trips on frequency based schedules
**Describe the problem**

`date` and `trip_id_performed` are necessary but not sufficient to uniquely identify a trip performed, which may require a trip `start_time` if the GTFS feed contains frequencies.txt.

For example:

**GTFS trips.txt**
| route_id | service_id | trip_id | direction_id | shape_id |
|---|---|---|---|---|
| r1 | weekday | trip1 | 0 | r1p1 |

**GTFS frequencies.txt**
| trip_id | start_time | end_time | headway_secs | exact_times |
|---|---|---|---|---|
| trip1 | 8:00 | 16:00 | 600 | 1 |

`trip1` is any of the trips on route `r1`, direction 0, service_id `weekday` that start between 8:00 and 16:00 (40 trips).

**Possible Solutions**

- Recommended: add `start_time` to the fields and primaryKey of trips_performed, making it required.
- or, make `start_time` conditionally required if GTFS contains frequencies.txt. No change to the primary key.
- or, require that when exact_times is true, frequencies.txt trips are expanded into the stop_times table. But this doesn't work when exact_times is false.

**Update: 2022-04-19**
The consensus is to continue to use the `trip_id_performed` and `service_date`, but to update the description of `trip_id_performed` to clarify that it doesn't need to be the same as the scheduled trip_id, and that it needs to be unique within a `service_date`. E.g., two or more vehicles performing the same scheduled trip would all have different values of `trip_id_performed`.

gabriel-korbato:
Update the description of trips_performed.trip_id_performed to clarify its the meaning and use.  Resolves #51 .

### DIFF
--- a/spec/trips_performed.schema.json
+++ b/spec/trips_performed.schema.json
@@ -20,7 +20,7 @@
     {
       "name": "trip_id_performed",
       "type": "string",
-      "description": "Uniquely identifies the trip performed.",
+      "description": "Uniquely identifies the trip performed. It must me unique for a service_date (and it can optionally be unique across all time). It does not have equal trip_id_scheduled.",
       "constraints": {
         "required": true
       }
--- a/spec/trips_performed.schema.json
+++ b/spec/trips_performed.schema.json
@@ -20,7 +20,7 @@
     {
       "name": "trip_id_performed",
       "type": "string",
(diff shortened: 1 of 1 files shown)


gabriel-korbato:
Fixed typos.

jlstpaul (approved):

PR #156, opened 2023-07-12
131 specify foreign key relationships that already exist but arent listed explicitly

### LINKED ISSUE #131: 📄🚀 – Specify foreign key relationships that already exist but aren't listed explicitly

As a software developer likely to use and produce TIDES data, I notice some internal foreign key relationships that aren't specified in the spec but that could be helpful to add.

I'd like to add these foreign keys to the spec.

**Additional context and sample data**

While working in issue #112 I noticed that `fare_transactions`, `vehicle_locations`, and `passenger_events` can sometimes or always (depending on the table) reference a `stop_visit` using the fields `(service_date, trip_id_performed, trip_stop_sequence)`. There may be other foreign key relationships that already exist and aren't specified but I haven't done an exhaustive check.

gabriel-korbato:
Partially resolves #131 .  This adds the foreign references that @jaygordon listed as examples, but I have not thoroughly reviewed the repo to find other missing references.  In the TIDES discussion from 2 weeks ago I believe we said that we might be fine saying that this is good enough for the 1.0 milestone, and that other missing references can be added if and when they are discovered.

A question for those more familiar with frictionless: am I following the correct format for a multi-column reference?

This commit works well if we assume that `trip_id_performed` is unique each day *and* that fare transactions from some service date can only correspond to stop visits from the same service date, which is true in most cases.  But there could be an agency that runs through the night, with vehicles simultaneously operating trips of different service dates at round 3 am, and in that case a fare transaction could be from a different service date than the stop visit.  This isn't currently handled by TIDES.

### DIFF
--- a/.github/workflows/validate_samples.yml
+++ b/.github/workflows/validate_samples.yml
@@ -4,11 +4,11 @@ on:
   push:
     paths:
       - 'samples/*/TIDES/*'
-      - 'spec/*'
+      # - 'spec/*' (TODO: Reinstate once samples folder exists)
   pull_request:
     paths:
       - 'samples/*/TIDES/*'
-      - 'spec/*'
+      # - 'spec/*' (TODO: Reinstate once samples folder exists)
   workflow_dispatch:
   create:
 
--- a/spec/fare_transactions.schema.json
+++ b/spec/fare_transactions.schema.json
@@ -184,10 +184,10 @@
       }
     },
     {
-      "fields": "stop_sequence",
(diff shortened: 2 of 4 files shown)


SorenSpicknall:
Hey @gabriel-korbato, if you merge in changes from `main` into this branch, you should get validation updates that'll resolve the OpenSSL bug on the currently failing Action.

SorenSpicknall (approved):

Issue #158, opened 2023-07-14
📄🚀 – Include vehicle location information from passenger count and fare transaction records


jlstpaul:
**Describe the feature you want and how it meets your needs or solves a problem**

As a creator of TIDES data, I want capture the location of passenger count and fare transaction events

The current spec does not include location information (lat/long, GPS quality, etc.) for `passenger_events` or `fare_transactions` records.  If an AFC or APC system includes location data, then the correct application of TIDES would be to create both a `passenger_events` or `fare_transaction` record _and_ a `vehicle_locations` record.

Question 1: Does this still sound right?  Or should we add location fields to the `passenger_events` and `fare_transactions` records?

Question 2: If we leave the fields as they are, do we want to be more explicit about how two records that represent the same event are related to each other. 

For example, if I have  single APC record that has both passenger counts and a location, I would create both a `passenger_events` record and a `vehicle_locations` record.  I want to be able to link those.  

Option 1: allow and require that the unique identifier for the two records match. So if the `passenger_event_id` and the `location_ping_id` are the same, that means the records are for the same event.

Option 2: add an optional reference to the a `vehicle_locations` record from the `passenger_event`s and `fare_transactions` tables.  

My preference is currently Option 2.

**Describe the solution you'd like**
Add field `passenger_events.location_ping_id` to identify the location of a passenger count
Add field `fare_transactions.location_ping_id` to identify the location of a fare transaction

**Describe alternatives you've considered**
Allow `passenger_events.passenger_event_id` and `fare_transactions.transaction_id` to match `vehicle_locations.location_ping_id` and define this to mean that the records are associated with the same event.

jlstpaul:
If there are no objections, I'd appreciate if someone could create a PR to implement Option 2.  This would add the following fields.  This should also include any necessary references to foreign keys

Add field `passenger_events.location_ping_id` to identify the location of a passenger count
Add field `fare_transactions.location_ping_id` to identify the location of a fare transaction

jlstpaul:
Closed in PR #164

PR #164, opened 2023-10-02
158 vehicle location information

jlstpaul:
Addresses #158 adding location_ping_id fields to passenger_events and fare_transactions.


jlstpaul:
@SorenSpicknall @e-lo can you tell me what I’m doing wrong here.

SorenSpicknall:
> @SorenSpicknall @e-lo can you tell me what I’m doing wrong here.

The error you're seeing isn't your fault - a file used for validation was deleted in a previous PR, and the relevant test just didn't happen to run after that until you drafted and submitted these changes. I've opened #165 to address the missing file.

spec/fare_transactions.schema.json
```diff
@@ -32,6 +32,12 @@
       }
     },
     {
+       "name": "location_ping_id",
+       "type": "string",
+       "title": "ID referencing vehicle_locations.location_ping_id",
+       "description": "Identifies a vehicle_location that is the location where the fare_transaction occurred"
```
botanize, line 38: I don't think you want to use "vehicle_location" when you mean a generic "vehicle location". Maybe `Identifies a vehicle location associated with this fare transaction`?

spec/passenger_events.schema.json
```diff
@@ -31,6 +31,12 @@
         "required": true
       }
     },
+    {
+      "name": "location_ping_id",
+      "type": "string",
+      "title": "ID referencing vehicle_locations.location_ping_id",
+      "description": "Identifies a vehicle_location that is the location where the passenger_event occurred"
```
botanize, line 38: See comment in fare_transactions.schema.json

botanize (changes requested):
Hi John,

I suggest simplifying the language used to describe `location_ping_id` to avoid confusion between table names and observations.

spec/fare_transactions.schema.json
jlstpaul: Agreed.  Thanks.

spec/passenger_events.schema.json
jlstpaul: Agreed. Thanks.

KatrinaMKaiser (changes requested):
Learned from @SorenSpicknall today that you'll also need to update the sample CSVs

botanize (approved):

KatrinaMKaiser (approved):

PR #162, opened 2023-07-22
Sample Data: document, update scripts to validate, update tests

e-lo:
# Pull Request

This Pull Request composes the necessary documentation and tests to develop, validate, and navigate TIDES data samples.

Documentation:
1. Documents sample data structure in `samples/README.md` (which then is added to the `/samples` page of documentation)
2. Documents available samples in `samples.md` (which becomes the `/samples` page of the documentation)
3. Documents the structure of a tides data package using a macro which reads `tides-datapackage-profile`
4. Updates `CONTRIBUTING.md` documentation 

Code: 
1. Adds a very lightweight script in `samples/template/scripts/create_template_files.py` which updates the template files based on the fields in the various table schemas. **note: this isn't run automagically anywhere b/c I think this wasn't a universally desired feature**
2 Add validation script `/bin/validate-datapackage`: validate a datapackage and its contents based on profile/spec specified in the datapackage, a local spec, or a remote spec based on its github reference.
3. Add validation script `/bin/validate-datapackage-to-profile`:  validate a datapackage file to the tides datapackage profile   based on profile/spec specified in the datapackage, a local spec, or a remote spec based on its github reference.
4. Splits `tests/test_all` into three scripts that we may want to use with different behaviors: 
  - `test_local_spec`  validates the local spec, documentation, and code
  - `test_samples_to_canonical` validates the local samples to the canonical spec and 
  - `test_samples_to_local` validates local samples to local spec

Spec:
1. Updates `tides-datapackage-profile` table schemas to be a string location to be consistent with most other profiles rather than an object.
2.  Updates `tides-datapackage-profile` examples to make it easier to use in documentation.
3. Removes `table-schema.json` b/c I don't think we needed it

---
Replaces https://github.com/TIDES-transit/TIDES/pull/100 which was mistakenly from a fork...
---

- [x] Pull-requests must address an existing issue
- [x] Update relevant documentation.

By contributing to this project, all contributors certify to the Developer Certificate of Origin in [CONTRIBUTING.md](CONTRIBUTING.md#contributor-agreement).

PR #165, opened 2023-10-03
Restore local table-schema.json

SorenSpicknall:
# Pull Request

In #149, we implemented a very slightly modified local copy of the standard `table-schema.json` file to support the `draft-07` JSON schema spec after testing various validation actions and finding a somewhat inelegant solution that allowed us to continue to validate what we wanted. This file [was inadvertently judged as inessential](https://github.com/TIDES-transit/TIDES/pull/162/commits/343060f031c0f246d9366c99d9601b9105b45d89) in #162 and deleted, and the deletion is now causing failures of [this GitHub action](https://github.com/TIDES-transit/TIDES/actions/workflows/validate_table_schemas.yml) that relied on the file.

This PR restores the deleted file to the repo.

### DIFF
--- a/spec/table-schema.json
+++ b/spec/table-schema.json
@@ -1,0 +1,1678 @@
+
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Table Schema",
+    "description": "A Table Schema for this resource, compliant with the [Table Schema](/tableschema/) specification.",
+    "type": [
+      "string",
+      "object"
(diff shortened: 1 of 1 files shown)


e-lo (approved):
LGTM but can you add an issue that would tidy this all up at some point ?

SorenSpicknall:
> LGTM but can you add an issue that would tidy this all up at some point ?

Will do!

PR #181, opened 2023-11-08
More robust link-updating

### LINKED ISSUE #170: 🐛💻 – broken links in samples/README.md, absolute links instead of relative
**Describe the problem**

There are two links defined in samples/README.md that don't work, they're both in here: https://tides-transit.github.io/TIDES/main/samples/#data-package in the bullet list under "See:"

The first one is defined as `{{site-url}}/datapackage`, and it should probably link to docs/datapackage.md, the other is `https://raw.githubusercontent.com/TIDES-transit/TIDES/main/samples/template` and should probably be `https://raw.githubusercontent.com/TIDES-transit/TIDES/main/samples/template/TIDES/datapackage.json`.

There are some links to the correct destinations in `/README.md`, but those specify a full path in the main branch, e.g., `https://raw.githubusercontent.com/TIDES-transit/TIDES/main/samples/template/datapackage.json`. It seems like documentation for each pull-request should link to references within that branch, not back to main.

**Steps to reproduce**

e-lo:
Fixes #170 

- Moves links to reference lists to make them easier to change. 
- Makes links relative to branch you are on - doesn't hard code to `main`
- Updates find/replace to use reference lists
- Makes sure that including file sections also includes references
- Testing/fixing of references/links working


main.py
```diff
@@ -10,18 +10,80 @@
...
+# targets for these link keys will be updated upon doc build
+UPDATE_LINKS = {
+    "[architecture]":"./architecture",
+    "[table schemas]":"./tables",
+    "[contributors]":"./development.md#contributors",
+    "[contributors.md]":"./development.md#contributors",
+    "[contributing]": "./development",
+    "[code of conduct]": "development#code_of_conduct",
+    "[CLA]":f"{GITHUB_REPO}/CLA.md",
+    "[license]": f"{GITHUB_REPO}/LICENSE)",
```
botanize, line 25: is there supposed to be a `)` in this format string?

main.py
```diff
@@ -10,18 +10,80 @@
...
-    "(CODE_OF_CONDUCT.md)": "(development/#tides-code-of-conduct)",
-    "(CLA.md)": "(development/#tides-contributor-license-agreement)",
-    "CONTRIBUTING.md)": "development/)",
-    "<LICENSE>": "[LICENSE](https://github.com/TIDES-transit/TIDES/blob/main/LICENSE)",
-    "contributors.md)": "development/#contributors)",
-    "architecture.md)": "architecture)",
-    "tables.md": "tables",
-    "https://tides-transit.github.io/TIDES/governance/":"governance.md"
+BRANCH_NAME_KEYWORD = "{branch_name}"
+GITHUB_REPO = f"http://github.com/TIDES-transit/TIDES/tree/{BRANCH_NAME_KEYWORD}"
```
botanize, line 14: Why do we need `GITHUB_REPO`? Shouldn't `./LICENSE` work instead of `https://github.com/TIDES-transit/TIDES/tree/main/LICENSE`? Or does that not work?

botanize (commented):
I like the reference lists, but I don't understand the need for `replace_links_in_markdown`, if these are all relative links, and they are, shouldn't it just work? Why does CLA.md need to be replaced with the full path to the repo and branch instead of just letting GH resolve the relative link?

e-lo:
This is bc some docs like readme, CLA, contributors and contributing exist
in base GitHub folders rather than docs...but get read in and used in docs
so we don't have to duplicate info.

I agree it isn't ideal. I did do templating with the GitHub repo...

jlstpaul (approved):

PR #191, opened 2023-11-30
Fix implementation challenges (#187)

### LINKED ISSUE #187: 🐛💻 – Implementation Challenges: Foreign Key Issues and Style Inconsistencies
**Describe the problem**
After seeing TIDES had hit v1-dev, I tried implementing a full TIDES schema in our Oracle database and ran into a few inconsistencies/issues (mostly around foreign_keys):

**Bugs**

1. The primary key for `trips_performed` is `service_date` and `trip_id_performed`. Most foreign keys on `trip_id_performed` don't reference `service_date`, which is an issue for databases that require a Foreign Key to be a Primary Key in the reference table (Oracle and Postgres both require this at least).
   1. Could be fixed by adding `service_date` to the trip ID performed foreign key for any table where it exists.
   2. This includes `fare_transactions`, `passenger_events`, `stop_visits`, `vehicle_locations`
2. In `passenger_events`, the foreign key on `train_car_id` references a `train_car` table instead of `train_cars`
3. In `passenger_events`, the `stop_times` foreign key references `scheduled_stop_sequence` in `passenger_events`, but `stop_sequence` in `stop_times`. I think this should reference `stop_visits` instead of `stop_times`, and should reference `stop_sequence` in both tables instead?
4. In `stop_visits`, I wasn't quite sure what the the `scheduled_stop_sequence` foreign key was meant to account for, but both Oracle and Postgres break with the current setup. I think we'd have to make the foreign key match on `scheduled_trip_id` and a `service_date`/`schedule_date` as well? We don't include `scheduled_trip_id` in this table, personally I would recommend removing the foreign key and keeping the constraint at just > 0.

**Style**

1. The options in the enum for `device_type` all start with lower-case letters, while the first letter is capitalized in all other enums in TIDES. This doesn't really matter for usage, but it stood out to me.
  
2. The `vehicles` table has `capacity_bike`, `capacity_seated`, and `capacity_standing`, but has `wheelchair_capacity`. In `train_cars`, it has `capacity_seated`, `capacity_standing`, `bike_capacity`, and `wheelchair_capacity`
    1. It may be too late, but I would personally love to change `bike_capacity` to `capacity_bike` in `train_cars`. Could also consider changing `wheelchair_capacity` to `capacity_wheelchair` in both tables, but that's less noticeable since it's consistent.

**Steps to reproduce**
I would be happy to share all the PLSQL DDL I put together to create the tables, but I don't think it's possible to make minimally reproducible without access to an Oracle database.

Danishswag:
Fixes #187 

**Bugs**
- Added `service_date` to `trips_performed` foreign key in `fare_transactions`, `passenger_events`, `stop_visits`, `vehicle_locations`
- Changed `train_car` to `train_cars` in foreign key in `passenger_events`
- Pointed `passenger_events.trip_stop_sequence` in passenger_events to `stop_visits.trip_stop_sequence` in the `stop_visits` foreign key
- Removed foreign key to `stop_times`

**Styles**
- Modified enums for `devices` to start with an uppercase letter
- Changed all capacity columns to `capacity_xxxx`

### DIFF
--- a/samples/template/TIDES/train_cars.csv
+++ b/samples/template/TIDES/train_cars.csv
@@ -1,1 +1,1 @@
-train_car_id,model_name,facility_name,capacity_seated,wheelchair_capacity,bike_capacity,bike_rack,capacity_standing,train_car_type
+train_car_id,model_name,facility_name,capacity_seated,capacity_wheelchair,capacity_bike,bike_rack,capacity_standing,train_car_type
--- a/samples/template/TIDES/vehicles.csv
+++ b/samples/template/TIDES/vehicles.csv
@@ -1,1 +1,1 @@
-vehicle_id,vehicle_start,vehicle_end,model_name,facility_name,capacity_seated,wheelchair_capacity,capacity_bike,bike_rack,capacity_standing
+vehicle_id,vehicle_start,vehicle_end,model_name,facility_name,capacity_seated,capacity_wheelchair,capacity_bike,bike_rack,capacity_standing
--- a/spec/devices.schema.json
+++ b/spec/devices.schema.json
@@ -39,12 +39,12 @@
       "description": "Indicates the type of device.",
       "constraints": {
         "enum": [
-          "farebox",
-          "fare gate",
-          "fare vending machine",
-          "passenger counter",
-          "vehicle location",
-          "other"
+          "Farebox",
+          "Fare gate",
+          "Fare vending machine",
+          "Passenger counter",
+          "Vehicle location",
+          "Other"
         ]
       }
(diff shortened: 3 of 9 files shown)


e-lo:
@Danishswag in order to run the CI activities you need to do it form a branch of `TIDES-transit/TIDES` instead of your local fork.  You should be able to do this via [quick git push to origin](https://stackoverflow.com/questions/805222/how-to-copy-a-local-git-branch-to-a-remote-repo) from your local fork:  `git push origin 187-implementation challenges`

PR #201, opened 2023-12-06
add remove stale branches gh-action

### LINKED ISSUE #200: 🚀💻 – Remove stale branches with a github action
As a developer, I'd like stale branches to be removed regularly so that the branch lists and documentation list isn't so enormous to navigate.

e-lo:
Will mark branches as stale after 30 days of inaction and remove them 7 days later unless action is taken.

- Currently implemented as a dry-run so no deletions will be made until we are happy with the config.
- Must be implemented in `main` branch to work.

Fixes #200

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,11 +20,14 @@ jobs:
         fetch-depth: 0
     - name: Set github reference as Slug environment variable
       run: |
+        echo "github.head_ref: "${{ github.head_ref }};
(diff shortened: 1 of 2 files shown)


jlstpaul:
Can a branch that is part of an open pull request become stale?   Would that cause an issue for tasks that are just taking a long time, but are still progress?

e-lo:
On second thought, i'm going to close this right now....i just removed them manually and think that is probably fine for us right now

PR #206, opened 2024-01-06
Add info from one-pager

e-lo:
- add text from one-pager in readme.md which will also render to homepage
- also added minutes page for board meetings.

This is a non-normative change.


jlstpaul:
Thank you for adding this @e-lo !

The one part I am worried about is the mermaid diagram. It may beg more questions than it answer at this point. I'd rather leave this off for now and work toward more complete visuals in the best practices documentation.

e-lo:
@jlstpaul - i updated the readme in the PR to not have mermaid diagram.

jlstpaul (approved):

PR #212, opened 2024-01-13
Update issue templates to be consistent with governance.

### LINKED ISSUE #208: Update github templates based on change management policy
As a developer, I'd like to have a clear path forward to getting a change approved that is consistent with the TIDES Change Management Policy

Specifically:

- [ ] Normative Changes
- [ ] Non-Normative Spec Changes
- [ ] Version Releases
- [ ] Version pre-releases

e-lo:
- Adds steps for developing solutions for urgent and normative changes
- Adds more details about processes
- Differentiates between normative and non-normative changes
- Adds release and pre-release issue templates

Fixes #208


chrisyamas:
@e-lo checking in here as well. Since this pre-dates the governance updates that the Board approved last week (now in #281), the templates would need a refresh against that language once it merges.

So we could fold the template updates into the post-#281 cleanup as a fresh PR, or refresh this branch once #281 lands. Wondering if you and @jlstpaul (whose review request is still standing here) have any preference?

PR #214, opened 2024-01-13
add mermaid diagram with git graph diagram showing branching model

### LINKED ISSUE #213: Add diagram to show branching model
As a developer I want to be able to look at a git branching diagram rather than reading a bunch of text to understand it.

e-lo:
Fixes #213


e-lo:
This PR adds this diagram to change management documentation. @jlstpaul can/should we approve and merge? 
image

chrisyamas:
@e-lo sorry we never answered your earlier question here. I think this one still looks useful (and it's small) so if we did a quick refresh against current docs, mark it ready, then it can merge through the normal docs update process? I'm happy to handle that refresh if that helps, or you should feel free to take it since you kicked this one off. @jlstpaul any objection to moving this one forward?

jlstpaul:
This is fine with me. Thank you for reviving it to resolve it @chrisyamas and thank you for working on the details @e-lo

PR #215, opened 2024-01-18
Change requirement for cleaning to be from TIDES repo.

### LINKED ISSUE #210: 🐛  CI not deleting refs in mike when refs to branches deleted
Expected:

[`clean-docs.yml`](https://github.com/TIDES-transit/TIDES/blob/main/.github/workflows/clean-docs.yml) workflow should be deleting the documentation version using [mike's delete functionality](https://github.com/jimporter/mike?tab=readme-ov-file#deleting-docs)

Instead:

...When I deleted branches via the GitHub web UI, the workflow didn't run.

e-lo:
Previous requirement didn't work and nothing was cleaned.

Fixes #210

### DIFF
--- a/.github/workflows/clean-docs.yml
+++ b/.github/workflows/clean-docs.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   clean:
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    if: github.repository_owner == 'TIDES-transit'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2


SorenSpicknall (approved):
We can't test this easily without recreating the whole docs publication flow elsewhere, so for convenience's sake (since the workflow is already not doing what it's supposed to) I'm supportive of merging this and then testing with a couple of sample deletions in prod. If the new criteria doesn't pick up our intended deletion prompt, we won't be any worse off than we are now. Approved!

e-lo:
Well...success on it running the action - not success on it working. @SorenSpicknall that looks scary...it was trying to delete `main` !?



image

SorenSpicknall:
There's a couple of things happening here. No risk to the branches themselves, but the specific pathway for a branch deletion was based on the context of the Action run after it was triggered by a deletion event. This meant that the `head_ref` at that time was `main` instead of the name of the deleted branch.

In #216, I changed the ref to use the branch from the deletion event itself, rather than the state of the repo after the deletion. The warnings about branch/repo state may just have to do with the attempt to operate on the `main` branch, but we'll see - we might need to do a little bit of cleanup of the automated `gh-pages` branch to get things fully operational after #216 is merged.

PR #216, opened 2024-01-18
Fix up event ref for deletion of docs associated with deleted branch

SorenSpicknall:
# Pull Request

Changes the branch lineage referenced in `clean-docs.yml` so that older docs versions associated with a branch's deletion event are targeted correctly for deletion, rather than the docs associated with the head branch after that event has happened. This is a follow-up to #215.

The event trigger behavior and the resulting branch ref was tested on a private repo before creating this PR.

### DIFF
--- a/.github/workflows/clean-docs.yml
+++ b/.github/workflows/clean-docs.yml
@@ -17,7 +17,7 @@ jobs:
       if: github.head_ref == ''
       run: |
         IFS="\/";
-        read -a strarr <<< "${{ github.ref }}";
+        read -a strarr <<< "${{ github.event.ref }}";
         last_ref="${strarr[${#strarr[*]}-1]}";
         echo "Slug=$(echo $last_ref)" >> "$GITHUB_ENV"
     - name: Document which reference


e-lo:
Ahhh - `event.ref` instead of `ref` :-) let's see how this goes...

e-lo:
no go :-(
image

SorenSpicknall:
This is progress. It's targeting the right branch for docs deletion now, and the remaining error is a different issue (probably just a bad state in the `gh-pages` branch). I'll get some time to look at that later today.

SorenSpicknall:
Because of the longstanding incomplete firing of documentation-related CI tasks, the `gh-pages` branch was in an unworkable state to be interacted with by the cleaning workflow. I deleted the branch and recreated it fresh from `main`.

@e-lo, it appears that the deletion automatically reset the branch source for docs publication. I don't have direct access to the settings needed to set that source again. Can you do the following for me?

1. Go to the repo Settings page
2. Click on the "Pages" section in the sidebar
3. Under "Build and deployment - Source - Deploy from a branch", select the `gh-pages` branch as the publishing source

Let me know when that's done, and I'll continue some testing.

e-lo:
Bad state 🤔 very interesting...

Did what you asked:

image

SorenSpicknall:
I'm guessing we also need to point the source at /docs instead of at the root. Can you do that?

e-lo:
It should deploy to root not docs

SorenSpicknall:
That was speculative - looking closer, it looks like top-level should be correct since we've got `mkdocs.yml` and `main.py` at the top level. I wasn't around when this was configured, so my apologies that I didn't inspect the setup closely enough when firing off that suggestion.

I think we've possibly done conflicting things while both trying to remedy this simultaneously, which is why the docs site isn't yet up. If we switch the source ref back to the root folder, and then run the full pages build and deployment process, hopefully that'll bring us back to where we should be.

SorenSpicknall:
Katrina switched the ref back this morning. I'll continue to test and keep you posted!

SorenSpicknall:
@e-lo, I've confirmed that [the deletion action now functions correctly](https://github.com/TIDES-transit/TIDES/actions/runs/7616621066), and that the docs section creation process also works upon a new PR from a repo branch - I triggered the creation action anew on #212 and #214 by making small copy changes to those PRs.

I believe the only remaining current need is to reset the default reference used by Mike to redirect people who visit the root of https://tides-transit.github.io/ (or https://tides-transit.org/) to the proper version of the docs. I wasn't sure what was used as a default before, so I didn't set one unilaterally. Currently, all active repo branches are present in the Mike selector, in addition to a version of the docs corresponding to the `v1.0-beta.1` tag. Let me know if you need any support setting the default to whatever is appropriate!

e-lo:
fixed!

PR #223, opened 2024-10-02
📄  [DOCS] Add Meeting Note Links

e-lo:
Added overall link to rolling agendas/meeting as well as links to heading of previous and future mtgs.

### DIFF
--- a/docs/governance.md
+++ b/docs/governance.md
@@ -34,7 +34,7 @@ TIDES is collectively owned and governed by a Board of Directors.
 
 ### TIDES Board of Directors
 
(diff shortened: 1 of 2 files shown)


jlstpaul:
@e-lo Two modifications requested;
1. In https://tides-transit.org/update-agenda-link/governance/meetings/ , remove the list of board meetings and just link to the rolling agenda.  Then we won't have to always update the page after each meeting.
2. On https://tides-transit.org/update-agenda-link/governance/#tides-board-of-directors at the top of the BOD section, change the link to the update rolling agenda.

Issue #224, opened 2025-01-10
🚀💻 – Change TIDES acronym to Transit Integrated Data Exchange Specification


jlstpaul:
**Describe the feature you want and how it meets your needs or solves a problem**
As a stakeholder of **TIDES**, I want the acronym for **TIDES** to be relevant, complete, and memorable so that everyone easily understands what TIDES is and what it for.

The current acronym is **Transit ITS Data Exchange Specification**.  **ITS** in turn is an initialism for Intelligent Transportation Systems. ITS is no longer a common term in the transit industry.  Therefore it does not help TIDES to be a relevant and memorable acronym.

**Describe the solution you'd like**
I propose that we change the acronym of TIDES to Transit **Integrated** Data Exchange Specification. This is more relevant and more memorable. It is not necessarily more complete.

**Describe alternatives you've considered**
Another alternative would be to use a different word for the letter "I" of the TIDES.
Or we could change the TIDES acronym altogether.

**Additional context**
This issue was discussed briefly at the TIDES Update Meeting on January 8, 2025 and there was general agreement then that 
a) the TIDES acronym should be retained, and 
b) the acronym meaning should be changed to Transit Integrated Data Exchange Specification.  

I can't even begin to think about how many files need to change to reflect this change.  I share my deep appreciation in advance to whomever prepares the PR for this...

PR #225, opened 2025-02-10
Update acronym: ITS --> Integrated

e-lo:
Per issue #224

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,4 +1,4 @@
-site_name: "Transit ITS Data Exchange (TIDES) Data Specification Suite"
+site_name: "Transit Integrated Data Exchange (TIDES) Data Specification Suite"
 site_url: https://TIDES-transit.github.io/TIDES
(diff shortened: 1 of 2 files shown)


botanize (commented):
I like this.

I found two more uses of "ITS", in spec/fare_transactions.schema.json ("ITS device") and docs/governance.md ("ITS vendor"). If ITS is not longer standard terminology in the industry perhaps those should be updated as well?

mkdocs.yml
```diff
@@ -1,4 +1,4 @@
-site_name: "Transit ITS Data Exchange (TIDES) Data Specification Suite"
+site_name: "Transit Integrated Data Exchange (TIDES) Data Specification Suite"
```
jlstpaul, line 1: Would you be okay with: "Transit Integrated Data Exchange Specification (TIDES) Suite"

e-lo:
@jlstpaul  - just updated per your apt suggestion. Please re-review

jlstpaul (approved):

Issue #235, opened 2025-07-10
🐛📄 – Improve event_type naming convention in passenger_events


chrisyamas:
## Problem Description

The current TIDES spec uses event types as strings with spaces (e.g., "Passenger boarded") in the `passenger_events` table. This naming convention creates practical challenges when working with the data, particularly in:

- Column naming for unpivot operations
- Variable naming in code and queries
- General data manipulation tasks

## Current Specification

The `passenger_events` table includes event types such as:
- "Passenger boarded"
- "Passenger alighted"
- "Vehicle arrived at stop"
- "Ramp deployed"
- "Kneel deployed"

These string values with spaces cannot be directly used as column names or variable names in most programming languages and database systems without modification.

## WMATA's Implementation Experience

During WMATA's implementation:

- Internal event types are represented as numeric codes (e.g., event_type values 3 and 5)
- Conversion between internal codes and TIDES string values requires additional mapping logic
- Analytical operations that involve pivoting or unpivoting data by event type become more complex
- Variable naming in code requires additional transformation of the event type strings

While this issue is relatively minor compared to other implementation challenges, it creates unnecessary friction in working with TIDES data.

## Possible Solution

We propose changing the event type naming convention to use underscore-delimited strings instead of spaces:

| Current Format | Proposed Format |
|----------------|-----------------|
| "Passenger boarded" | "passenger_boarded" |
| "Passenger alighted" | "passenger_alighted" |
| "Vehicle arrived at stop" | "vehicle_arrived_at_stop" |
| "Ramp deployed" | "ramp_deployed" |
| "Kneel deployed" | "kneel_deployed" |

This change would maintain human readability while significantly improving technical usability.

### Option 1: Provide Both Code and Description

Another approach would be to follow the pattern used in other TIDES enumerations, providing both a code and a description:

```json
{
  "event_type_code": "passenger_boarded",
  "event_type_description": "Passenger boarded"
}
```

### Option 2: Maintain Current Format with Implementation Guidance

If changing the specification is not desirable, provide implementation guidance on how to handle spaces in event types for common operations:

```
When using event types in column names or variable names, implementers should replace spaces with underscores and convert to lowercase.
```

## Impact

Adopting underscore-delimited strings would:

1. Improve usability in database operations (pivoting, unpivoting, filtering)
2. Simplify code that references event types
3. Reduce the need for string manipulation in implementations
4. Align with common naming conventions in data engineering

## Backward Compatibility

[...]

PR #246, opened 2025-12-29
Add vehicle_crew supporting table

### LINKED ISSUE #220: 🐛📄 – TODS and TIDES Alignment: Mid-Trip Reliefs
**Describe the problem**
TODS allows multiple runs per trip (particularly common on commuter rail, but also happens on bus service at CTA, NYCT, and WMATA), while TIDES specifies one operator in the `trips_performed` table. Much less important than ridership/runtime analysis, but supporting mid-trip reliefs and perhaps additional staff could enable other types of analyses using TIDES formatted data.

**Possible Solutions**
1. Move `operator_id` to the `stop_visits` table.
2. Specify that `operator_id` is always the first operator on the trip.
3. Remove `operator_id` from `trips_performed` and create a new `trips_performed_crew` table to support additional multiple operators and additional staff (e.g. conductors)

For 3, a possible format could be something like the following table, though it probably needs some refinement:

| `trip_id_performed` | `employee_id` | `role` | `start_time` | `end_time` |
| - | - | - | - | - |

**Additional context**
TODS v2 is moving to a format where it is a supplement to GTFS. Since TIDES was built assuming some agencies would want to reference GTFS fields, most fields should be compatible, and it should be straightforward to build TIDES as if non-revenue information in TODS is just in the normal schedule.

TODS v2 isn't quite final yet, though the working group has voted to move to the GTFS supplement format.

chrisyamas:
## Summary

Adds new `vehicle_crew` table to support multiple crew members per vehicle/trip, including mid-trip reliefs. This addresses the limitation where `trips_performed.operator_id` could only reference a single operator.

Resolves #220

## Changes

- `spec/vehicle_crew.schema.json` (new) defines the `vehicle_crew` table with:
  - Primary key: `[service_date, trip_id_performed, crew_id, start_time]`
  - Required fields: `service_date`, `vehicle_id`, `crew_id`, `start_time`
  - Optional fields: `trip_id_performed` (for vehicle-level assignments), `crew_role`, `end_time`
  - `crew_role` enum: `operator`, `conductor`, `fare_inspector`, `other` (uses snake_case consistent with #235)
  - Foreign references to `trips_performed`, `vehicles`, and `operators`
- `samples/template/TIDES/vehicle_crew.csv` (new) provides header-only template
- `spec/trips_performed.schema.json` updated `operator_id` description to clarify optional usage and reference to `vehicle_crew` table for multi-operator scenarios
- `spec/tides-datapackage-profile.json` added `vehicle_crew` to `tides-table` enum
- `samples/template/TIDES/datapackage.json` added `vehicle_crew` resource entry
- `CHANGELOG.md` added entries under `[Unreleased]` for new table and description change

## Reason for this change

Rail systems commonly have multiple crew members (operators, conductors) on a single trip, and operator reliefs mid-trip are standard practice. The current `trips_performed.operator_id` field can only reference one operator, forcing agencies to either lose crew data or pick an arbitrary operator to record.

The new `vehicle_crew` table allows agencies to capture complete crew assignment history while keeping the existing `operator_id` field for simple single-operator cases.

_- See related discussion in [Issue #220](https://github.com/TIDES-transit/TIDES/issues/220) and Fall 2025 TIDES Issues Working Group notes from [December 1](https://docs.google.com/document/d/1vjd7788aC0Ua4tySMl7ytcDjSV_lUMXi4FPV_-c3ArM/edit?usp=sharing) and [December 10](https://docs.google.com/document/d/1cIYkhlEscs1wpUZz8D2USQr8EYJNQg5NDeUJQHtN2gw/edit?usp=sharing)._

## Review checklist

Per [change management policy](https://tides-transit.org/main/governance/policies/change-management/#normative-content), the following must be met before feature branch changes can merge to `develop` branch:

- [x] All JSON files validate
- [ ] Reviewed and approved by 2+ contributors or board members

---

### Community review status (updated September 2, 2026)

**How community review works:** Per the [TIDES Change Management Policy](https://tides-transit.org/main/governance/policies/change-management/#__tabbed_1_4), a proposal moves forward once at least three TIDES Contributors outside the originating working group publicly comment with a score: **Accepted**, **Accepted with minor changes**, or **Substantially revised**. A few sentences with your read of the proposal is a complete review.

**Originating working group (Fall 2025 Issues Working Group, December 1 and 10, 2025 sessions):** Christopher Yamas, Gabriel Sánchez Martínez, Ian Thistle, Jay Gordon, John Levin, Joey Reid, Spenser Sutinen. Their work is reflected in the proposal itself; community review comes from Contributors beyond this group.

**Review so far:** Laurie Merrell (@laurie-jarvus) and @jabhij have posted scored reviews (Accepted; Accepted with minor changes, with clarifications answered in the thread). One more review from a Contributor outside the working group completes community review.

**Community review remains open, and reviews are welcome while we finalize v2.0. If you have not yet weighed in, a short comment closing with where you land (**Accepted** / **Accepted with minor changes** / **Substantially revised**) is all we need.**

The full v2.0 map lives on #271.


chrisyamas:
> **Data Validation Report**
> 
> Sample	Status
> ./samples/template/TIDES	⚠️

This warning is because the frictionless validation check can't find the `vehicle_crew.schema.json` since it checks against `main` branch and the schema for the new table doesn't exist on `main` branch yet. But this should resolve automatically once the PR is merged.

jlstpaul (approved):

spec/trips_performed.schema.json
botanize, line 1: It looks like you've accidentally removed `trip_id_performed` from trips_performed!

spec/vehicle_crew.schema.json
```diff
@@ -0,0 +1,99 @@
...
+      "description": "Service date. References GTFS",
+      "constraints": {
+        "required": true
+      }
+    },
+    {
+      "name": "trip_id_performed",
+      "type": "string",
+      "title": "ID referencing trips_performed.trip_id_performed",
+      "description": "Identifies the trip performed. Optional if crew assignment is at vehicle level only."
```
botanize: Shouldn't `trip_id_performed` be required if it's part of the primary key?

CHANGELOG.md
```diff
@@ -26,9 +34,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `vehicle_locations` - Timestamped vehicle locations and speeds.
     - `passenger_events` - Timestamped passenger-related events, including boardings and alightings.
     - `fare_transactions` - Timestamped fare transaction, associated with devices.
-    - `stop_visits` - Summarized boarding, alighting, arrival, departure, and other events (kneel engaged, ramp deployed, etc.) by trip and stop for each service date.
+    - `stop_visits` - Summarized boarding, alighting, arrival, departure, and other events (kneel engaged, ramp deployed, etc.) by trip and stop for each service     date.
     - `trips_performed` - Trips performed for each service date.
-    - `station_activities` - Summarized transactions, entries, and exits by stop or station and time period for each service date (for events not associated with a trip).
+    - `station_activities` - Summarized transactions, entries, and exits by stop or station and time period for each service date (for events not associated with a     trip).
```
botanize: Looks like there's some extra whitespace in these lines.

botanize (changes requested):
I believe this meets the needs discussed in the issue. However, it looks like this PR accidentally removes `trip_id_performed` from the `trips_performed` table. I also suspect that `trip_id_performed` should be required if it's part of the primary key. Though to be fair, it doesn't strictly need to be, service_date, crew_id and start_time should be sufficient for the primary key. Including trip_id_performed just makes it easier to use.

- [ ] undo removal of `trips_performed.trip_id_performed`

botanize (approved):

jaygordon (approved):

laurie-jarvus (approved):
Accepted

jabhij:
**Accepted with minor changes**

I think this is a useful addition and addresses the core limitation identified in #220: `trips_performed.operator_id` cannot adequately represent multiple crew members or mid-trip reliefs. The use of `start_time`/`end_time` also gives the model the temporal granularity needed for relief scenarios.

I have a few points that I think are worth clarifying before finalizing the schema:

1. **`crew_id` vs. `operators`** -- `crew_role` allows `operator`, `conductor`, `fare_inspector`, and `other`, but `crew_id` currently references `operators.operator_id`, while the `operators` table is defined specifically as personnel who operate vehicles. If `vehicle_crew` is intended to represent all crew, should the referenced entity be broader than `operators`, or should the specification explicitly require all crew members to be represented in `operators` regardless of role?

2. **Vehicle/trip consistency** -- because both `vehicle_id` and `(service_date, trip_id_performed)` are stored, the schema can theoretically contain a `vehicle_crew` record whose `vehicle_id` differs from `trips_performed.vehicle_id`. If that is intentional to support vehicle changes during a trip, I think the documentation should state that explicitly. If not, it would be useful to define the expected consistency relationship.

3. **Temporal semantics** -- since mid-trip relief is a primary use case, I would recommend documenting the expected temporal rules for `start_time`/`end_time`, particularly whether `end_time` is exclusive, whether `end_time >= start_time` is required, and how overlapping crew assignments should be interpreted.

_(Kindly consider these mostly as semantic/integrity clarifications rather than objections to the overall approach.)_

jlstpaul:
Thank you @jabhij for the great comments/questions.  Here are my thoughts:

> * **`crew_id` vs. `operators`** -- `crew_role` allows `operator`, `conductor`, `fare_inspector`, and `other`, but `crew_id` currently references `operators.operator_id`, while the `operators` table is defined specifically as personnel who operate vehicles. If `vehicle_crew` is intended to represent all crew, should the referenced entity be broader than `operators`, or should the specification explicitly require all crew members to be represented in `operators` regardless of role?

We should probably update the name of the `operators` table to `crew` table. And we should add a `crew_type` field to that table.  I appreciate others' thoughts on that. (And it will probably require `start_date` and `end_date` fields so that individual crew members can change roles over time but retain the same identifier.).  To be honest, the current `operators` table has really just been a placeholder that we knew was going to be expanded upon eventually.

> * **Vehicle/trip consistency** -- because both `vehicle_id` and `(service_date, trip_id_performed)` are stored, the schema can theoretically contain a `vehicle_crew` record whose `vehicle_id` differs from `trips_performed.vehicle_id`. If that is intentional to support vehicle changes during a trip, I think the documentation should state that explicitly. If not, it would be useful to define the expected consistency relationship.

My understanding is that a trip that is performed by multiple vehicles (e.g., vehicle switch or doubled up) should be represented by multiple `trips_performed` records.  In that case, there should be a rule that `trips_performed.vehicle_id` is the same as `vehicle_crew.vehicle_id` for the same trips or trip segments.

> * **Temporal semantics** -- since mid-trip relief is a primary use case, I would recommend documenting the expected temporal rules for `start_time`/`end_time`, particularly whether `end_time` is exclusive, whether `end_time >= start_time` is required, and how overlapping crew assignments should be interpreted.

I don't have an opinion on whether `end_time` is exclusive.  My assumption would be that it is.  I do think that `end_time` should be >= `start_time`.  

Does overlapping crew assignments mean that same `crew_id` is found on two different vehicles at the same time?  That probably shouldn't be allowed...

jabhij:
Thank you @jlstpaul for the thoughtful response. I agree with the direction, especially treating `operators` as a placeholder that can evolve into a broader `crew` entity.

A few follow-up thoughts:

1. For the proposed `crew_type` in the broader `crew` table, I think it may be useful to distinguish that from `vehicle_crew.crew_role`. My interpretation would be that `crew_type` describes the person's classification/role in the crew master data, while `crew_role` describes the role they performed for that particular vehicle/trip assignment. If that distinction is intended, documenting it could help avoid ambiguity and potential redundancy.

2. I agree with representing a vehicle switch/doubling as separate `trips_performed` records. In that case, it may be useful to explicitly document the integrity relationship that `vehicle_crew.vehicle_id` must match the corresponding `trips_performed.vehicle_id` for the same performed trip/segment. It would also be helpful to clarify whether `trip_id_performed` is intended to uniquely identify that vehicle-specific performed trip/segment.

3. On the temporal side, I agree that `end_time >= start_time` should be required, and I would also lean toward treating `end_time` as exclusive. For overlapping assignments, I think the rule may need to be defined at the `crew_id` level. For example, the same crew member should generally not be assigned to two different vehicles during overlapping intervals, whereas overlapping records for the same vehicle may be legitimate depending on the crew roles. Clarifying this distinction would make the temporal behavior easier for implementers to interpret consistently.

Overall, I think these clarifications would make the intended relationships and temporal semantics much clearer without changing the core approach.

mpaine-act:
I noticed the use of the word crew while [TODS](https://tods-transit.org/spec/#dataset-files) uses employee.  Does this make any difference to the spec?

gabriel-korbato:
@mpaine-act We considered "employee" and other words, but settled for "crew" because (1) it is shorter, (2) it is more specific to employees that perform functions in vehicles and stations (not administration), and (3) it could include contractors that are technically not employees.

jlstpaul:
> 1. For the proposed `crew_type` in the broader `crew` table, I think it may be useful to distinguish that from `vehicle_crew.crew_role`. My interpretation would be that `crew_type` describes the person's classification/role in the crew master data, while `crew_role` describes the role they performed for that particular vehicle/trip assignment. If that distinction is intended, documenting it could help avoid ambiguity and potential redundancy.

@jabhij It was not my intent to differentiate `crew_type` from `crew_role`, so I would prefer to use the same `crew_role` field name if we do create a `crew` table.

chrisyamas:
heads up @jlstpaul @botanize @jaygordon @laurie-jarvus @jabhij that this proposal has been revised in response to the community review we've had here over the past few weeks. The `operators` table is now a `crew` table (`crew_id`, `crew_role`, `start_date`, `end_date`), following Abhishek's questions about the `crew_id` referent and temporal fields and John's operators-to-crew direction, and `vehicle_crew`, `trips_performed`, and `vehicle_train_cars` now reference `crew.crew_id`.

in accordance with the [change management policy](https://tides-transit.org/main/governance/policies/change-management/#__tabbed_1_4), the revised proposal is re-circulating for at least 72 hours with previous reviewers notified; but if nothing here gives you pause, it will proceed (fresh reviews are welcome either way!)

spec/trips_performed.schema.json
```diff
@@ -229,12 +229,12 @@
     {
       "name": "schedule_trip_start",
       "type": "datetime",
-      "description": "Scheduled departure time from the trip’s origin."
+      "description": "Scheduled departure time from the trip\u2019s origin."
```
wtimmerman-fitp, line 232: nit: replace with ascii apostrophe throughout?

spec/trips_performed.schema.json
```diff
@@ -167,7 +167,7 @@
           "Light Rail",
           "Monorail/Automated Guideway",
           "Other",
-          "Público",
+          "P\u00fablico",
```
wtimmerman-fitp, line 170: re: other nit on unicode, this seems fine

spec/vehicle_crew.schema.json
```diff
@@ -0,0 +1,108 @@
+{
+  "primaryKey": [
+    "service_date",
+    "trip_id_performed",
+    "crew_id",
+    "start_time"
+  ],
```
botanize, line 7: The goal seems to be to track the crew on a vehicle. I'm surprised to see trip_id_performed included here, but not vehicle_id. 

What happens to an interlined block, or a single end-to-end trip that changes route mid-run? For example, TriMet's Orange/Yellow lines are interlined and change route in downtown Portland. 

My understanding is that no crew changes typically occur when the route changes. Including trip_id_performed in the primary key would be generating a new set of crew records for the same vehicle crew, simply due to interlining. 

Much simpler would be to use vehicle_id in the primary key instead of trip_id_performed, and make the trip id optional. You can easily enough match crew to a trip by using the vehicle of the performed trip and the start and end times.

spec/vehicle_crew.schema.json
```diff
@@ -0,0 +1,108 @@
...
+        "required": true
+      }
+    },
+    {
+      "name": "trip_id_performed",
+      "type": "string",
+      "title": "ID referencing trips_performed.trip_id_performed",
+      "description": "Identifies the trip performed. Required as part of the primary key.",
+      "constraints": {
+        "required": true
```
botanize, line 28: see above, but I don't think this should be required.

botanize (changes requested):
I feel pretty strongly that the vehicle_crew table should not include trip_id_performed in the primary key and the field should be made optional. Instead use vehicle_id.

jabhij:
Thanks @botanize for raising the interlining example. After reviewing the relationship between `vehicle_crew` and `trips_performed`, I think this highlights an important question about the intended grain of `vehicle_crew`.

I agree that making `trip_id_performed` part of the primary key could cause the same continuous crew assignment to be represented as multiple records simply because the vehicle moves from one performed trip to another. This seems particularly relevant for interlined/block service, where TIDES already allows multiple sequential `trips_performed` records to use the same vehicle.

At the same time, I would hesitate to treat `vehicle_id` as a direct replacement for `trip_id_performed` without explicitly defining the grain of the table.

I see two possible interpretations:

1. **Vehicle-level crew assignment:** a crew member is assigned to a vehicle for a time interval, potentially spanning multiple performed trips.
2. **Trip-level crew assignment:** a crew member is explicitly associated with a particular performed trip, which can be useful for mid-trip relief and trip-level analysis.

If the intended grain is the first case, I agree that the primary key should be based on the vehicle-level assignment; potentially `service_date + vehicle_id + crew_id + start_time` with `trip_id_performed` optional when that relationship is available from the source system.

This also seems workable because `trips_performed` already contains `vehicle_id`, `actual_trip_start`, and `actual_trip_end`, allowing a vehicle-level crew interval to be associated with the relevant performed trips when needed.

I would still keep `trip_id_performed` as an optional field rather than remove it entirely. If an agency has an explicit trip-to-crew relationship, preserving that identifier gives us stronger referential integrity than relying solely on temporal matching.

Perhaps the key question to resolve is therefore: **Is `vehicle_crew` intended to represent a crew assignment to a vehicle over a time interval, or an assignment to an individual performed trip?**

If it is the former, I agree with the concern about `trip_id_performed` being part of the primary key and would favor making it optional.

jlstpaul:
> Perhaps the key question to resolve is therefore: Is vehicle_crew intended to represent a crew assignment to a vehicle over a time interval, or an assignment to an individual performed trip?

I think the answer here is "both".  The original intent of this change was to address the situation of reliefs when there were multiple operators on the same vehicle for the same trip. In that case there would be just one `trips_performed` record and thus only one `operator_id` field.   This is why `trip_id_performed` is prominent in the original PR.   

At the same time, I think we are discovering that a more pure table linking vehicles and crew, independent of trip details, is desirable.  

I think that Joey's proposal, to remove `trip_id_performed` from the primary key and make it optional and instead include `vehicle_id` in the primary key allows us to do both. **In the normal course of events, every `vehicle_crew` record will likely have a value for the optional `trip_id_performed` field.  This allows for easy joins withs the `trips_performed` table**, which was the original intent.  But in some cases, or with some raw data sources, it will be allowable not to have a `trip_id_performed` value.

chrisyamas:
Heads up @jlstpaul @botanize @jaygordon @laurie-jarvus @jabhij @wtimmerman-fitp @mpaine-act that this proposal has been revised again in response to the review here, so it's re-circulating per the change management policy (at least 72 hours, previous reviewers notified).

With this round of changes, the `vehicle_crew` primary key is now `[service_date, vehicle_id, crew_id, start_time]`, anchored on the vehicle rather than the trip, and `trip_id_performed` is optional (usually populated, but a single assignment can span consecutive trips). These changes have followed Joey's point about interlining, Abhishek's question about the grain of the table, and John's identification of the value of keeping the trip reference available for easy joins. The `crew` table from the previous round remains unchanged.

If nothing here gives you pause, it will proceed, and fresh reviews are welcome either way. We still need one more scored review from outside the original working group on this one.

jlstpaul (approved):

PR #257, opened 2026-04-30
Update TIDES governance page: current Program Manager affiliations

chrisyamas:
Updates the Program Manager row on the governance page to reflect the current assignment per TIDES Board direction. Removes the two prior co-manager entries and lists Christopher Yamas (MobilityData Contractor) as the current TIDES Program Manager.


chrisyamas:
The failing `build-deploy` check is a pre-existing CI/toolchain issue, not caused by this change.

Root cause was that `.github/workflows/docs.yml` sets `python-version: '3.x'`, which now resolves to Python 3.14. Python 3.14 dropped `pkg_resources` from the default install, and `docs/requirements.txt` pins `mike<2.0`, which still imports `pkg_resources` on startup. Any PR running through this workflow will fail until the docs build dependencies include `setuptools` (or until `mike` is upgraded / Python is pinned).

Fix in flight as a separate PR: #258.

jlstpaul (approved):

Issue #173, opened 2023-11-03
🐛💻  mike rebase no longer supported 


e-lo:
**Describe the problem**

Per mike [changelog](https://github.com/jimporter/mike/blob/076a4af3270a448f6aeb880c9c6c2fc0d80f603f/CHANGES.md?plain=1#L30), the --rebase command is no longer supported in v2.0+

**Steps to reproduce**

Submit a pull request that runs github actions to deploy documentation.

botanize:
The temporary pinning is now three years old, and not so temporary, see PR #258. It may now be causing some minor issues, for example, we must also pin setuptools to version < 81, which leads to some dependabot security notifications that probably don't affect this project, but could be resolved by modernizing the workflow.

It looks like the easiest alternative to `mike --rebase` is to run `git checkout gh-pages && git reset --hard origin/gh-pages`, as noted here: https://github.com/jimporter/mike/issues/196#issuecomment-1900977999.

What's not clear to me is why we needed `--rebase` in the first place. The creator of mike suggests this usually happens when you've done something else wrong, but we're just checking out the branch in CI.

PR #275, opened 2026-08-10
fix(deps): upgrade setuptools to >=83 and resolve dependency chain breakage

fredericsimard:
# Description

GitHub security alert flagged `setuptools <83.0.0` in requirements.txt ([CVE-2026-59890](https://www.cve.org/CVERecord?id=CVE-2026-59890), moderate severity). See original email screenshot below.

Upgrading `setuptools` to `>=83` surfaced several downstream breakages, resolved here:

- setuptools >=83 removes `pkg_resources`; pin `mike>=2.0` (was <2.0), since `mike 1.1.2` depended on `pkg_resources`
- pin `pyyaml>=6.0.1` to avoid a build failure against `setuptools 83`
- remove unused `jsonschema-cli` dependency in `/requirements.txt`, which pinned `pyyaml<6.0` and blocked the pyyaml upgrade
- pin `setuptools>=83,<84.0.0` in `/docs/requirements.txt`

Verified with `mkdocs serve`, builds and serves locally after these changes with zero visible issues or bugs.

# README additions

Added instructions on how to serve and build the site locally, tested and functionning.

# Pull Request

Before submitting a pull request, please read [CONTRIBUTING.md](CONTRIBUTING.md).

For spec changes:

By contributing to this project, all contributors certify to the Developer Certificate of Origin in [CONTRIBUTING.md](CONTRIBUTING.md#contributor-agreement).

# Original email

Screenshot 2026-08-10 at 16 18 22

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,5 @@
 -r docs/requirements.txt
+frictionless
+jsonschema
+jsonschema-cli
+pre-commit
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,1 @@
 -r docs/requirements.txt
-frictionless
-jsonschema
-jsonschema-cli
-pre-commit
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Build docs
       run: |
         git fetch origin gh-pages --depth=1
-        mike deploy --push --rebase ${{ env.Slug }}
+        mike deploy --push ${{ env.Slug }}
         echo "Documentation available at: http://tides-transit.github.io/TIDES/${{ env.Slug }}"
   comment:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -72,6 +72,28 @@ Those who want to help with the development of the TIDES specification should re
 
 Please add issues, bugs, and feature requests to [GitHub][github issues].
 
+## Building or Serving the site locally
+
+1. In Terminal, change the directory to where you would like the local clone of this repo to be installed.
+1. Ensure you have an up-to-date version of pip: 
+   - Linux: `pip install pip` or `pip install --upgrade pip`
+   - macOS: `pip3 install pip` or `pip3 install --upgrade pip`
+1. Clone this repository:
+   - `git clone https://github.com/TIDES-transit/TIDES.git`
+1. Change the directory to the cloned repository, and create & enable a Python virtual environment:
+   - `python3 -m venv venv`
+   - `source venv/bin/activate`
+1. Install dependencies:
+   - `pip3 install --force-reinstall -r requirements.txt`
+1. To run the site locally:
+   - `mkdocs serve`
+   - Then go to this address with any browser: `http://127.0.0.1:8000/TIDES/`
+1. To build the site locally only:
+   - `mkdocs build`
+   - the files will end up in the `/site` folder.
+1. Deactivate the Python virtual environment when done:
+   - `deactivate`
+
 ## Acknowledgment
 
 These data schemas and tool definitions developed here are based on the results of research conducted by the Transportation Research Board (TRB) of the National Academies of Science, Engineering, and Medicine (NASEM) under the Transit Cooperative Research Program (TCRP). This research is available at [the National Academies website][tcrp report].
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-mike<2.0
+mike>=2.0
 mkdocs
 mkdocs-autorefs
 mkdocs-awesome-pages-plugin
@@ -9,6 +9,6 @@ mkdocs-material
 mkdocs-mermaid2-plugin
 mkdocs-redirects
 pandas
-pyyaml
-setuptools<81
+pyyaml>=6.0.1
+setuptools>=83,<84.0.0
 tabulate


fredericsimard:
> Documentation available at: [`http://tides-transit.github.io/TIDES/fix/setup-tools-vuln/2026-08-10`](http://tides-transit.github.io/TIDES/fix/setup-tools-vuln/2026-08-10)

FYI: this link is 404 - Not found.

botanize:
The workflow still uses the removed `--rebase` flag, https://github.com/TIDES-transit/TIDES/blob/8183f089f257eaeb41d98544ddf70e66c29bb6f8/.github/workflows/docs.yml#L49. See #173

fredericsimard:
> The workflow still uses the removed `--rebase` flag,
> 
> https://github.com/TIDES-transit/TIDES/blob/8183f089f257eaeb41d98544ddf70e66c29bb6f8/.github/workflows/docs.yml#L49
> 
> . See #173

@botanize What does `mike deploy` do? Is this a replacement for `mkdocs serve` ? Never seen that before.

And should I update the workflow to remove `--rebase`?

botanize:
My understanding is the `mike deploy` pushes a version of the docs to gh-pages, and allows multiple versions to coexist (e.g., docs for different branches or tags). You have to remove `--rebase` as it's not supported in mike >=2.0. The suggestion, if you need, it is to run `git checkout gh-pages && git reset --hard origin/gh-pages`. Though you might just try removing the `--rebase` flag first and see if that's needed.

fredericsimard:
@botanize All checks successful, that was it.

requirements.txt
```diff
@@ -1,5 +1 @@
--r docs/requirements.txt
-frictionless
-jsonschema
-jsonschema-cli
-pre-commit
```
botanize, line 5: I think these might be required for other workflows, can you check the contents of `.github/workflows`?

botanize (commented):
I'm a little worried about removing the dependencies from requirements.txt as they might be used by other workflows. Can you confirm that these dependencies are not used by other workflows?

I found the compiled documentation for this pr here: https://tides-transit.org/fix-setup-tools-vuln-2026-08-10/, which is not the link provided by the workflow in the PR. Would be good to get that fixed, but that could be another PR.

requirements.txt
fredericsimard: They were failing the `mkdocs serve` and `mkdocs build`, they seemed to be incompatible. One thing we can do is skip the `requirements.txt` at the root and only use the one in `docs/`. That I know works just fine.

fredericsimard: Ok the two workflows that update the site use the `docs/requirements.txt` so no danger here. The `requirements.txt` at the root is used by some other scripts but not workflows: [results of search for `requirements.txt` in repo](https://github.com/search?q=repo%3ATIDES-transit%2FTIDES+requirements.txt&type=code).

I think we can revert the changes I made to the root `requirements.txt` file but it might still break the few scripts who call on it as itself calls on the `docs/requirements.txt` and might break them due to incompatibilities.

There's a good number of things that need updating: old versions of tools in workflows, deprecation warnings, incompatibilities when some tools are updated...

We could just table this one for now, but a medium vulnerability is not something to ignore long-term.

fredericsimard: @botanize Do you agree with my suggestion above?

botanize: Ok, I reviewed all of the files that call the top-level requirements.txt and none of them are actually run. They're all examples. But, I'm surprised the top-level requirements.txt dependencies were causing issues with mkdocs serve. I just ran mkdocs serve after `uv add -r requirements.txt` and everything worked fine, though it looks lke mkdocs might be nearing its end (see #276). Can you revert changes to the top-level requirements file and we can see how CI/CD handles it?

fredericsimard: As far as I understand it the CI/CD skips the requirements.txt at the root of the repo, goes straight to the docs/requirements.txt and as such will bypass the issues entirely after I bring back the plugins. Other scripts (not workflows/actions) somewhere may break though.

mkdocs 1.x is going to be replaced by mkdocs 2.x sometime in the future and it's a major breaking change, no sites will work when updated to it, apparently. ProperDocs is a drop-in replacement (a fork of 1.x) which promises no breaking change and continued improvements. I will before the end of the year move all tech sites at MobilityData to the latter.

Materials for Mkdocs though is no longer getting updates and they created a new project called Zensical which in the long term may prove to be an excellent replacement for Materials, but I decided to go with MaterialX (a fork of Materials) which is a perfect drop-in replacement with minimal changes and improvements. Most sites are using this new skin, all of them will by the end of the year also.

fredericsimard: Both checks have completed successfully after discarding the changes to the root requirements.txt file.

botanize (approved):

fredericsimard:
@botanize Oddly it says "Merging is blocked" even though all requirements are satisfied. Do you want me to use the "Merge without waiting for requirements to be met (bypass rules)" option?

botanize:
Might require approval or merging by someone in a particular team, @chrisyamas can you provide some insight?

chrisyamas:
> Might require approval or merging by someone in a particular team, @chrisyamas can you provide some insight?

Yes, it's because it's merging directly into main branch. It should have approval by Board reviewers before merge into main. I would suggest @jlstpaul and @gabriel-korbato to review.

jlstpaul (approved):

fredericsimard:
> > Might require approval or merging by someone in a particular team, @chrisyamas can you provide some insight?
> 
> Yes, it's because it's merging directly into main branch. It should have approval by Board reviewers before merge into main. I would suggest @jlstpaul and @gabriel-korbato to review.

@chrisyamas @botanize Even with the latest approval, it still says Merging is blocked. Could it be that it needs more people to approve? The rule says only 1 reviewer is required though.

chrisyamas:
It turned out that the root cause wasn't the review requirements, as those are fully satisfied. The ruleset on `main` includes a restrict-updates rule that separates review approval from merge execution, such that while approvals (including code owners) satisfy review, only a 'bypass actor' (the Board, the Manager's repo admins, or an org admin) can execute the merge.

So the "Merging is blocked" warning reflects the execution gate. In practice this is how merges to `main` have worked all along, where approved PRs have the merge executed by an admin through the bypass control.

This definitely helped highlight that we've been missing clear documentation on this, so the banner has been seeming like a quirk (and a lack of maintenance on this, since the bypass list still referenced a team that was deleted years ago early in the repo's history).

I recommend we keep the execution gate model here, since it enforces the change management policy's separation between approval and release execution.

Re: maintenance on this, the old team has now been removed from the bypass list, the `TIDES Board` team has been added (in alignment with change management policy), and the bypass has been narrowed to PR-only, which closes the gap where admins could previously push commits to main directly without a PR. Now every change is ensured to have to go through an approved PR, and admins then needed for executing the merge.

The model is now documented in #277.

@fredericsimard, since approvals are satisfied and you're a bypass actor, this merge is yours to execute. The button will still be labeled as a bypass, as intended. Thanks @botanize and Fred for surfacing this!

fredericsimard:
Perfect, I'll bypass and merge. One suggestion is you could add 2 reviewers minimum and block bypass instead. That's what we do on our side.

